PR #133, opened 2015-10-04
Use boostrap form-group for upload-form

Thor77:
screenshots:
upload form
-------
binary upload form

### DIFF
--- a/bepasty/templates/index.html
+++ b/bepasty/templates/index.html
@@ -1,84 +1,80 @@
 {% extends "_layout.html" %}
 
+{% macro maximum_lifetime() -%}
+<div class="row">
+  <div class="col-md-3 form-group">
+    <input class="form-control" name="maxlife-value" type="number" min="1" value="1" /><br>
+  </div>
+  <div class="col-md-9 form-group">
+    <select class="form-control" name="maxlife-unit" size="1">
+      <option value="forever">Keep Forever</option>
+      <option value="minutes">Minutes</option>
+      <option value="hours">Hours</option>
+      <option value="days">Days</option>
+      <option value="weeks">Weeks</option>
+      <option value="months">Months</option>
+      <option value="years">Years</option>
+    </select>
+  </div>
+</div>
+{% endmacro %}
+
 {% block content %}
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
it is hard to see why this is an improvement, can you explain?
it is also hard to see how this is meant to look as a whole.

the text box and the dropzone that is usually on the left somehow belong together as they are both ways to submit single item(s).

the smaller text box with the id list and the list of clickable single-item links that usually appear on the right after uploading stuff also somehow semantically belong together.

Thor77:
Oh, sorry the screenshots were a bit missleading.
Before:
bepasty_before
After:
bepasty_after

ThomasWaldmann:
the wider spacing looks better, but some fields are unnecessarily wide now and the whole left part takes a lot of vertical space (and the old version already hardly fitted on screen height).

on the left side: content-type, filename and button are usually all rather non-wide elements, can you have them sit on one line (as before, just a little more spacing so it looks better)?

on the right side: there seems to be a bit much vertical space between filename and button (also: it is much more than on the left side). So the submit almost looks disconnected from the form it belongs to.

also: do you think it is necessary to have the extra field label texts outside of the field if we show them inside the field already? Takes vertical space of which we often do not have enough.

You removed the hint "choose before dragging or submitting", maybe better keep that?

Thor77:
bepasty_new
If you want to merge this, please let me squash the commits before.

ThomasWaldmann:
That looks pretty (thanks!), I'ld like to merge that after one last fix:

On the right side, you have now also that "(choose before dragging or submitting)" hint. But there is no dropzone on the right side (and it should be obvious there to do it before submit as it sits inside the form).

Thor77:
screenshot at 12-51-24
Done :)

ThomasWaldmann:
OK, so it is finished for merging?

Thor77:
I don't think you want all these commits in your commit-history, want you?

ThomasWaldmann:
If you like, you can clean up history, but it is not a requirement.

Thor77:
There you go, ready to merge :)

PR #135, opened 2015-10-05
bepasty-server.nix: implements build-support for nix package manager

makefu:
`bepasty-server.nix` defines the build inputs and steps to create a self-contained
version of the bepasty-server.
This provides an 'immutable' build environment for nix-users.

### DIFF
--- a/bepasty-server.nix
+++ b/bepasty-server.nix
@@ -1,0 +1,138 @@
+
+## usage:
+## Build the project with dependencies
+#      nix-build bepasty-server.nix -A bepasty-server
+## Install bepasty-server into environment
+#      nix-env -f bepasty-server.nix -iA bepasty-server
+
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Is it a problem to put this in another directory than project-root?
I'ld prefer not having stuff there that doesn't need to be there.

Maybe packaging/nixos/... ?

makefu:
Due to the fact that both the package is now in nixpkgs ( NixOS/nixpkgs/pull/10239 ) and the bepasty-server is an open pull-request ( NixOS/nixpkgs/pull/10512 ) i do not think this PR is needed anymore.
Maybe we could add a reference in the documentation that the project available in nixos.

ThomasWaldmann:
@makefu Hmm, I don't think we have a "available on" list. Do we need one? The users of x usually can just look in their package manager.

PR #137, opened 2015-10-06
Add Favicon

Thor77:
I'm not sure if `static/app/` is the right place for the svg-file, but the other folders doesn't rly match, too.


ThomasWaldmann:
isn't it 16x16 for non-highdpi displays?

Thor77:
[This](http://www.w3.org/2005/10/howto-favicon) W3-standard says
> However, the format for the image you have chosen must be 16x16 pixels or 32x32 pixels, using either 8-bit or 24-bit colors

ThomasWaldmann:
Well, the wikipedia page seems more interesting for practice here - e.g. it tells that png does not work for IE < 11.

And a a w3c "draft in development" from 2005 looks a bit strange anyway. :)

But we can just try as it is and improve later, if needed. Thanks for your effort!

PR #138, opened 2015-10-11
Use icons instead of text for permissions

Thor77:
permission_icons

### DIFF
--- a/bepasty/app.py
+++ b/bepasty/app.py
@@ -63,7 +63,7 @@ def create_app():
         stuff here and make it easily available.
         """
         flaskg.logged_in = logged_in()
-        flaskg.permissions = get_permissions()
+        flaskg.permissions = get_permission_icons()
         if flaskg.logged_in:
             session.permanent = current_app.config['PERMANENT_SESSION']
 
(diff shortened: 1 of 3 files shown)


ThomasWaldmann:
how about accessibility?

Thor77:
oh, i didn't think about it from that point of view.
Do you think some hover-texts for the icons are enough?
Or do you generally dislike the idea of using icons for permissions?

ThomasWaldmann:
iirc, img alt attrs are mandatory for that. hover texts are likely useful also.

pguth:
I think the icon selection is good. Alt-texts should suffice imho.

Thor77:
added title-attr to icons for hover-text.
it seems like the alt-attr doesn't have any affect for the i-tag: i tried viewing the page with `lynx` but neither the icons nor the alt-text appeared.

ThomasWaldmann:
yeah, alt might work only for img.

Thor77:
What's about this XStatic-Stuff (i'm currently just using boostrap-CDN)?
It seems like the [XStatic-Font-Awesome](https://github.com/stackforge/xstatic-font-awesome) package is quite outdated...

NotANormalNerd:
I am also not in the favour of those xstatic packages. The idea behind them is okay, but without the ecosystem it is just point blank useless. And the advantage is minor.

Thor77:
It had some advantages for security, if you didn't rly trust your CDN.
But since at least Firefox + Chrome support [Subresource Integrity](https://hacks.mozilla.org/2015/09/subresource-integrity-in-firefox-43/), that's not a problem anymore.

pguth:
Don't forget about the [`crossorigin` attribute](https://hacks.mozilla.org/2015/09/subresource-integrity-in-firefox-43/)! \o/

ThomasWaldmann:
crossorigin anonymous looks nice.

PR #142, opened 2015-10-22
Initial test-setup

Thor77:
* removed old test-files
* added `.travis.yml` to run the tests automatically
* removed now unused dependencies
* added coverage-checks

If you want to merge this, please activate [Travis](https://travis-ci.org/) and [coveralls](https://coveralls.io/) for this repository.


.travis.yml
```diff
@@ -0,0 +1,12 @@
+language: python
+python:
+  - "2.7"
+  - "3.4"
```
ThomasWaldmann: what about 3.3?

bepasty/tests/conftest.py
```diff
@@ -0,0 +1,34 @@
...
+import pytest
+
+from bepasty.app import create_app, create_storage
+
+
+@pytest.fixture(scope='module')
+def testclient(request):
+    '''
+    creates a Flask-testclient instance for bepasty
+    '''
```
ThomasWaldmann: please use triple-double quotes for docstrings, some tools have issues with triple-single.

bepasty/tests/conftest.py
```diff
@@ -0,0 +1,34 @@
...
+    app = create_app()
+    db_file, app.config['DATABASE'] = mkstemp()
```
ThomasWaldmann: py.test offers tmpfile and some other tmp related fixtures - can you use them?

bepasty/tests/conftest.py
```diff
@@ -0,0 +1,34 @@
...
+    # setup a secret key
+    app.config['SECRET_KEY'] = str(random())
+    # setup permissions
+    app.config['PERMISSIONS'] = {
+        'l': 'list',
+        'c': 'create',
+        'r': 'read',
+        'd': 'delete',
+        'a': 'admin'
+    }
```
ThomasWaldmann: each password just giving one permission is a bit unusual. but if it works for the tests...

bepasty/tests/conftest.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def teardown():
+        close(db_file)
+        unlink(app.config['DATABASE'])
+    request.addfinalizer(teardown)
```
ThomasWaldmann: have a look at yield_fixture, it's nice.

bepasty/tests/conftest.py
Thor77: do you have another/better suggestion?

ThomasWaldmann: having sets of permissions with incrementing power?
their names could be either something verbosely readable or just their sorted first letters combined.

ThomasWaldmann:
a general question: you killed a lot of tests without replacement, why did you do that?

Thor77:
I just *killed* all tests, because I want to have all tests in one location (`bepasty/tests/`) and generally thought they're deprecated. But it seems like at least some of them are still up2date, should i re-add all or just some of them?

ThomasWaldmann:
Well, maybe first try to run the tests before you kill them.
I am not aware that there were major problems with them.

bepasty/tests/conftest.py
Thor77: I already wrote one [permission-test](https://github.com/Thor77/bepasty-server/blob/permission_tests/bepasty/tests/test_permissions.py) and just wanted to have a simple way to test the permissions one-by-one, it wouldn't be a big problem to use permissions with incrementing power, but i just didn't need it.

bepasty/tests/conftest.py
Thor77: The `tmpdir`-fixture is [function-scoped](https://travis-ci.org/Thor77/bepasty-server/jobs/87001685#L230) (maybe there's still a way to use it, but due the lack of experience with py.test I've no idea) and the `tmpdir_factory`-fixture is [not available](https://travis-ci.org/Thor77/bepasty-server/jobs/87002362#L265) for the latest pytest-version for py2.7

Thor77:
Seems like all the tests are still working.
My [travis-builds](https://travis-ci.org/Thor77/Thor77/bepasty-server/builds/87019261) pass now \o/

Thor77:
Maybe you want to insert the travis and coveralls-badge into README?

ThomasWaldmann:
Yes, would be nice (at the end, so it is less annoying for people reading it with "less").

PR #144, opened 2015-10-27
Fix #143

### LINKED ISSUE #143: permissions don't work
after commit https://github.com/bepasty/bepasty-server/commit/734c64a870b92b28475db1d6ddbbc77743d054b0 permissions for read,create,... don't work.
I can't create any new pastes or access existing pastes.

Thor77:


bepasty/utils/permissions.py
```diff
@@ -54,7 +54,7 @@ def get_permissions():
 def get_permission_icons():
     return [
         (permission, permission_icons[permission])
-        for permission in get_permissions()
+        for permission in get_permissions() if permission
```
ThomasWaldmann: can you explain the "if permission" part?

Thor77: if you set no permissions, get_permissions will return [''] and because permission_icons has no key '', it will raise an IndexError

ThomasWaldmann: hmm, that should be rather the empty set (not  set with 1 empty-string member).
but ok, that bug is not in your stuff.

Thor77: feel free to try it yourself, maybe my setup was just misconfigured or smth like that

ThomasWaldmann: no, your code is ok I guess. didn't try, but I hope you did this time.

i was just referring to some other code that needs to get optimized so that such "if permission" code is not needed, because the set only contains valid members.

PR #145, opened 2016-10-04
Fix delete button in detail view

rixx:
The delete button did not work in the detail view of a paste (didn't even send a request).


ThomasWaldmann:
thanks!

PR #150, opened 2017-03-08
support running the app at non-root URL pathes

ThomasWaldmann:


qknight:
seems to work very well!
http://nixdoc.io/bepasty/Q7y4oxph

i've uploaded a picture and a textfile!
thanks so much!

ThomasWaldmann:
@qknight you're welcome. btw, first time I have practically seen that "don't login via insecure http" warning from firefox. did you already try https also?

PR #154, opened 2018-01-05
prevent int cast on empty string

foosinn:
Fixes file upload for me:

```
[2018-01-05 09:52:44,898] ERROR in app: Exception on /+upload/new [POST]
Traceback (most recent call last):
  File "*hidden*/env/lib/python3.5/site-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "*hidden*/env/lib/python3.5/site-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "*hidden*/env/lib/python3.5/site-packages/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "*hidden*/env/lib/python3.5/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "*hidden*/env/lib/python3.5/site-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "*hidden*/env/lib/python3.5/site-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "*hidden*/env/lib/python3.5/site-packages/flask/views.py", line 84, in view
    return self.dispatch_request(*args, **kwargs)
  File "*hidden*/env/lib/python3.5/site-packages/flask/views.py", line 149, in dispatch_request
    return meth(*args, **kwargs)
  File "./bepasty/views/upload.py", line 75, in post
    maxlife_value = int(data['maxlife_value'])
ValueError: invalid literal for int() with base 10: ''

```

### DIFF
--- a/bepasty/views/upload.py
+++ b/bepasty/views/upload.py
@@ -72,7 +72,7 @@ class UploadNewView(MethodView):
         data_type = data['type']
 
         # set max lifetime
-        maxlife_value = int(data['maxlife_value'])
+        maxlife_value = int(data.get('maxlife_value', 1))
         maxlife_unit = data['maxlife_unit'].upper()
         maxtime = time_unit_to_sec(maxlife_value, maxlife_unit)
         maxlife_timestamp = int(time.time()) + maxtime if maxtime > 0 else maxtime


bepasty/views/upload.py
```diff
@@ -72,7 +72,7 @@ def post(self):
         data_type = data['type']
 
         # set max lifetime
-        maxlife_value = int(data['maxlife_value'])
+        maxlife_value = int(data.get('maxlife_value', 1))
```
ThomasWaldmann: sorry, i somehow missed your PR and just found it by accident now.

the question here is: why is maxlife_value missing from post data?

maxlife_unit (see below) is access in the same way and if it does not crash there also, it seems to be always there.

ThomasWaldmann: also, your fix does not address "empty string", but rather a missing value (so it would be None, except when fixed like in your pull request).

ThomasWaldmann:
superceded by #186. thanks for finding this.

PR #187, opened 2018-10-06
[Issue #173] Fix flake8 errors and warnings

### LINKED ISSUE #173: fix flake8 (pep8) warnings/errors
in `setup.cfg`, there are currently some flake8 warnings/errors being excluded because they are triggered by the bepasty code.

remove 1 exclude, fix all failures, commit, repeat.

pywkm:
Resolves #173 [fix flake8 (pep8) warnings/errors]

No more global flake8 error ignoring. Few remarks:
- F821 - it's about assigning `xrange` to variable in Py3, so I can only disable this check locally (inline)
- E731 - after removing not used variables it applies only to two lambda statement - pretty clear and short, so no need for deifining functions - I also left noqa inline
- file `docs/source/conf.py` after one of recent changes (#162) tries to import `from bepasty import project, version, copyright_years, author, description`, which do not exist there anymore. It's not in scope of this issue (flake8 doesn't fail because of this), but should be fixed if you need to (re)generate Sphinx documentation
- as far as I know, changes from that PR didn't brake anything (I don't have pypy, Py3.4 and Py3.7 locally installed, so couldn't run complete tox suite). App was still working during manual tests. Unit tests are passing. Four selenium tests were failing, but not because my change - master was already broken.


src/bepasty/utils/_compat.py
```diff
@@ -4,21 +4,17 @@
...
-    iteritems = lambda d: iter(d.items())
-else:
     range_type = xrange
+    iteritems = lambda d: d.iteritems()  # noqa: E731
+else:
+    from urllib.parse import urlparse, urljoin
 
-    from urlparse import urlparse, urljoin
-    iterkeys = lambda d: d.iterkeys()
-    itervalues = lambda d: d.itervalues()
```
ThomasWaldmann: why did you remove these 2?

src/bepasty/utils/upload.py
```diff
@@ -128,6 +126,6 @@ def create_item(f, filename, size, content_type, content_type_hint,
 @threaded
 def background_compute_hash(storage, name):
     with storage.openwrite(name) as item:
-        size = item.meta[SIZE]
+        size = item.meta[constants.SIZE]
         file_hash = compute_hash(item.data, size)
-        item.meta[HASH] = file_hash
+        item.meta[constants.HASH] = file_hash
```
ThomasWaldmann: I see how you put `constants.` to avoid the pep warning, but i think the old way was easier to read.

If there is only one module where such uppercase constants come from, I think this is considered as a valid use of a star import.

So, just add noqa to silence pep8?

src/bepasty/views/delete.py
```diff
@@ -4,23 +4,22 @@
...
-from ..utils.permissions import *
+from .. import constants
+from ..utils import permissions
 from ..utils.http import redirect_next_referrer
 
 
 class DeleteView(MethodView):
     def post(self, name):
-        if not may(DELETE):
+        if not permissions.may(permissions.DELETE):
```
ThomasWaldmann: same/similar here. old was easier to read.

src/bepasty/utils/_compat.py
```diff
@@ -11,7 +11,7 @@
 if PY2:
     from urlparse import urlparse, urljoin
 
-    range_type = xrange
+    range_type = xrange  # noqa: F821
```
ThomasWaldmann: did you run the pep8 checker on py2 and py3? this is somehow asymetric now.

src/bepasty/apis/__init__.py
```diff
@@ -1,6 +1,13 @@
...
+
 blueprint = Blueprint('bepasty_apis', __name__, url_prefix='/apis')
 
-from . import (
-    lodgeit, rest)
+blueprint.add_url_rule('/lodgeit/', view_func=LodgeitUpload.as_view('lodgeit'))
+blueprint.add_url_rule('/rest', view_func=InfoView.as_view('api_info'))
+blueprint.add_url_rule('/rest/items', view_func=ItemUploadView.as_view('items'))
+blueprint.add_url_rule('/rest/items/<itemname:name>', view_func=ItemDetailView.as_view('items_detail'))
+blueprint.add_url_rule('/rest/items/<itemname:name>/download', view_func=ItemDownloadView.as_view('items_download'))
```
ThomasWaldmann: hmm, this is moving around quite a lot of stuff just to get flake8 silent.

i dislike a bit that the rules are not added in the module where the functionality is implemented, but everything is moved up to init.

src/bepasty/utils/_compat.py
pywkm: Because it was never used? I mean, not only in current master - I don't see if it was even used at the day of adding it to repository (in year 2014). Looks like too broad stack overflow copypasta.

src/bepasty/utils/upload.py
pywkm: It's not only silencing pep8 warnings. It's just ugly and confusing to use `from module import *`. You should avoid that in EVERY situation, you are polluting namespace with thing you are not aware of.
So if not `from .. import constants` and not `from ..constants import *`, maybe this?
```
from ..constants import (
    FOREVER,
    FILENAME,
    HASH,
    SIZE,
    ...    
)
```

src/bepasty/views/delete.py
pywkm: Importing everything (*) from `permissions` shows how bad it is. You didn't even realize that in `src/bepasty/apis/rest.py` the `request` and `current_app` weren't even imported directly form flask, but from `permissions` - that's why I had to add their imports.

src/bepasty/utils/_compat.py
pywkm: Yes I did. Of course it's asymetric, `range` is perfectly fine keyword in Py2. The other way - not.

src/bepasty/utils/upload.py
ThomasWaldmann: yeah, the latter would be a good way to solve both issues. bit harder to maintain, but guess we can live with that.

src/bepasty/apis/__init__.py
pywkm: Well, and I can say that it's strange to add them to ***blueprint*** in place other than definition of blueprint. Because of that (creating blueprint in one place and extending it in other place) you needed to split those imports to two parts to avoid circular imports. That was a signal, you had something wrong with project structure. IMO.

src/bepasty/utils/_compat.py
ThomasWaldmann: ah, right.

src/bepasty/apis/__init__.py
ThomasWaldmann: ok, you convinced me. :)

src/bepasty/utils/_compat.py
ThomasWaldmann: well, i was making the usual stuff available. iterkeys/values/items all fall into same class.

does keeping this trigger a flake8 issue?

pywkm: No, flake8 didn't complain it wasn't used. It was complaining about lambda expression though, so before I added noqa disabling this check (E731) I was curious where they are used.
It doesn't make sense to maintain unused chunks of code. It's OK to remove code, you know. ;) If you ever need that variables, git remembers them and restoring will be very easy.

ThomasWaldmann: ok.

PR #188, opened 2018-10-06
Fix selenium tests

pywkm:
As mentioned in neighboring pull request, selenium tests were failing on my machine. After short investigation, I'm not sure if this problem was indeed global. Maybe you didn't notice it at all. Thing is, after filling form test tries to find text on next page (view), which on my machine didn't even render yet. Adding small amount of sleep solves problem. 0.1s is enough for me, so I added 0.2s. ;)
Is it possible, that those test were passing everywhere else and that's why no one was reporting that issue earlier? I have little experience with selenium. :(


ThomasWaldmann:
I didn't run the selenium tests recently.

PR #205, opened 2019-05-22
Add support for asciinema recordings

### LINKED ISSUE #175: support asciinema player?

dasJ:
Just upload a *.cast file, and it will automatically render with the asciinema player.

Fixes #175.

### DIFF
--- a/src/bepasty/cli/server.py
+++ b/src/bepasty/cli/server.py
@@ -4,7 +4,6 @@ bepasty-server commandline interface
 
 from __future__ import print_function
 import argparse
-import mimetypes
 
 from ..app import create_app
 
@@ -17,9 +16,6 @@ def main():
 
     args = argparser.parse_args()
     app = create_app()
-
-    mimetypes.add_type('application/x-asciinema-recording', '.cast')
-
     print(" * Starting bepasty server...")
     app.run(host=args.host, port=args.port, debug=args.debug)
 
--- a/src/bepasty/cli/server.py
+++ b/src/bepasty/cli/server.py
@@ -4,6 +4,7 @@ bepasty-server commandline interface
 
 from __future__ import print_function
 import argparse
+import mimetypes
 
 from ..app import create_app
 
@@ -16,6 +17,9 @@ def main():
 
     args = argparser.parse_args()
     app = create_app()
+
+    mimetypes.add_type('application/x-asciinema-recording', '.cast')
+
     print(" * Starting bepasty server...")
     app.run(host=args.host, port=args.port, debug=args.debug)
 
--- a/src/bepasty/views/display.py
+++ b/src/bepasty/views/display.py
@@ -110,6 +110,10 @@ class DisplayView(MethodView):
                     src = url_for('bepasty.inline', name=name)
                     link_txt = u'Click to see PDF'
                     rendered_content = Markup(u'<a href="%s">%s</a>' % (src, link_txt))
+                elif ct == 'application/x-asciinema-recording':
+                    src = url_for('bepasty.download', name=name)
+                    rendered_content = Markup(u'<asciinema-player src="%s" title="%s" theme="%s" autoplay>' %
+                                              (src, item.meta[FILENAME], current_app.config['ASCIINEMA_THEME']))
                 elif use_pygments:
                     text = read_data(item)
                     # TODO we don't have the coding in metadata
--- a/src/bepasty/views/display.py
+++ b/src/bepasty/views/display.py
@@ -113,7 +113,8 @@ class DisplayView(MethodView):
                 elif ct == 'application/x-asciinema-recording':
                     src = url_for('bepasty.download', name=name)
                     rendered_content = Markup(u'<asciinema-player src="%s" title="%s" theme="%s" autoplay>' %
-                                              (src, item.meta[FILENAME], current_app.config['ASCIINEMA_THEME']))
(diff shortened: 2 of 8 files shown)


ThomasWaldmann:
thanks for the PR!

```
src/bepasty/views/display.py:116:40: E128 continuation line under-indented for visual indent
```
there's a formatting issue (pep8) that makes travis-ci fail.

also, i'ld like this to be based on a xstatic-package. if there is none yet and you don't want to make one yourself, i can make one in the next days.

dasJ:
@ThomasWaldmann Thanks for the review. I fixed the indentation according to flake8.

I can create a xstatic-package but I wouldn't going to maintain it, so you creating the package would probably make a lot more sense. Could you do that? I'll include it, then

ThomasWaldmann:
OK, I'll try to make a package next week. I'll ping back when finished.

dasJ:
Thanks!

ThomasWaldmann:
https://pypi.org/project/XStatic-asciinema-player/ <- please test it!

ThomasWaldmann:
ping!

dasJ:
Sorry for the long delay, I think it should be okay now

src/bepasty/utils/upload.py
```diff
@@ -66,6 +66,7 @@ def filter_type(cls, ct, ct_hint, filename=None):
         Only allow some basic characters and shorten to 50 characters.
         """
         if not ct and filename:
+            mimetypes.add_type('application/x-asciinema-recording', '.cast')
```
ThomasWaldmann: this doesn't look like the right place for this.

i suspect filter_type is a frequently called function and one does not need to call add_type more than once per process.

src/bepasty/views/display.py
```diff
@@ -110,6 +110,10 @@ def read_data(item):
                     src = url_for('bepasty.inline', name=name)
                     link_txt = u'Click to see PDF'
                     rendered_content = Markup(u'<a href="%s">%s</a>' % (src, link_txt))
+                elif ct == 'application/x-asciinema-recording':
+                    src = url_for('bepasty.download', name=name)
+                    rendered_content = Markup(u'<asciinema-player src="%s" title="%s" theme="%s" autoplay>' %
+                                              (src, item.meta[FILENAME], current_app.config['ASCIINEMA_THEME']))
```
ThomasWaldmann: can this be used for XSS by using strange filenames?

dasJ: I doubt it. In my tests, it used the generated filename: `/LdigCtA9/+download`

dasJ: Looks like bepasty handles that. I uploaded `demo.cast" lal.cast` which resulted in:
```
<asciinema-player src="/hFFgG6eb/+download" title="demo.cast lal.cast" theme="asciinema" autoplay="">
```

src/bepasty/utils/upload.py
dasJ: Yes, I can do that in cli

dasJ:
@ThomasWaldmann Fixed the location of `add_type`, hopefully this is better

src/bepasty/cli/server.py
```diff
@@ -16,6 +17,9 @@ def main():
 
     args = argparser.parse_args()
     app = create_app()
+
+    mimetypes.add_type('application/x-asciinema-recording', '.cast')
```
ThomasWaldmann: better, but still not the right place.

i guess this gets only invoked when starting the built-in server from flask, but not when using this with an external server (gunicorn, uswsgi, mod-wsgi) just as a wsgi app.

src/bepasty/views/display.py
ThomasWaldmann: hmm, ok, so it seems the filename given by the user gets preprocessed somewhere else.

src/bepasty/cli/server.py
ThomasWaldmann: did you see my comment?

src/bepasty/views/display.py
ThomasWaldmann: guess this should have a theme default like:
```
current_app.config.get('ASCIINEMA_THEME', 'asciinema')
```

ThomasWaldmann: bepasty preprocesses the filename before storing it into metadata and uses this regex (all matching chars will be stripped from filename):

```
class Upload(object):
    _filename_re = re.compile(r'[^a-zA-Z0-9 \*+:;.,_-]+')
```

src/bepasty/cli/server.py
ThomasWaldmann: check if this should be on top level of module, where mimetypes is first imported.
maybe right below the import statement.

ThomasWaldmann:
Note: i cleaned up some comments that are not relevant any more.

Stuff left is about TODO and about "no XSS here".

ThomasWaldmann:
@dasJ do you want to finish this or shall I?

src/bepasty/views/display.py
ThomasWaldmann: i fixed that myself.

src/bepasty/cli/server.py
ThomasWaldmann: fixed this myself.

ThomasWaldmann:
Thanks for the PR! :-)

PR #215, opened 2020-10-17
Fixes REST api, and add some new functionality

OGAWAHirofumi:
This PR does,

- fixes several bugs, adds tests, generalize error response for REST api
- 78dbb01: adds new REST api for delete/lock/unlock (for example, my own client uses those)
- 05976a4: allow to copy link for display etc. buttons in web site
- 3627356: optionally support python-magic to detect file type (this adds TYPE_HINT metadata, so can be arguable)

(resend: looks like something strange happens)

Thanks.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,6 +40,7 @@ setup(
     install_requires=[
         'Flask',
         'Pygments',
+        # 'python-magic',
         'xstatic',
         'xstatic-bootbox',
         'xstatic-bootstrap<4.0.0.0',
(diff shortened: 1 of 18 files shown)


ThomasWaldmann:
Thanks for the PR!

But, it is way too big and mixes all sorts of stuff.

Please break it down into multiple PRs with clean scope:
- on PR per idea
- different PRs for fixing existing stuff and adding new stuff

OGAWAHirofumi:
OK, I will do soon

PR #216, opened 2020-10-17
Fix REST api

OGAWAHirofumi:
This PR fixes the issues in REST api path and adds tests to prevent regression.

And REST api error response is consolidated to rest_errorhandler.error_message()
for consistency, and easier to test.

(there are 2 PRs that is depending on this)

### DIFF
--- a/src/bepasty/apis/rest.py
+++ b/src/bepasty/apis/rest.py
@@ -1,3 +1,4 @@
+import errno
 import base64
 import time
 from io import BytesIO
@@ -153,9 +154,17 @@ class ItemUploadView(RestBase):
             # Get file name from Transaction-ID and open from Storage
             trans_id_s = request.headers.get(TRANSACTION_ID)
             trans_id_b = trans_id_s if isinstance(trans_id_s, bytes_type) else trans_id_s.encode()
-            name_b = base64.b64decode(trans_id_b)
+            try:
+                name_b = base64.b64decode(trans_id_b)
+            except (base64.binascii.Error, TypeError):
+                raise BadRequest(description='Could not decode {}'.format(TRANSACTION_ID))
             name = name_b if isinstance(name_b, str) else name_b.decode()
-            item = current_app.storage.openwrite(name)
+            try:
+                item = current_app.storage.openwrite(name)
+            except (OSError, IOError) as e:
+                if e.errno == errno.ENOENT:
+                    raise BadRequest(description='Could not find storage item for transaction id')
+                raise
             new_item = False
 
         response = None
--- a/src/bepasty/apis/rest.py
+++ b/src/bepasty/apis/rest.py
@@ -4,7 +4,7 @@ from io import BytesIO
 
 from flask import Response, make_response, url_for, jsonify, stream_with_context, request, current_app
 from flask.views import MethodView
-from werkzeug.exceptions import HTTPException, Conflict, InternalServerError, MethodNotAllowed
+from werkzeug.exceptions import HTTPException, BadRequest, Conflict, Forbidden, InternalServerError, MethodNotAllowed
 
 from ..constants import FILENAME, ID, SIZE, TYPE, TRANSACTION_ID
 from ..utils._compat import bytes_type
@@ -57,19 +57,19 @@ class ItemUploadView(RestBase):
 
         # Check Content-Range. Needs to be specified, even if only one chunk
         if not request.headers.get("Content-Range"):
-            return 'Content-Range not specified', 400
+            raise BadRequest(description='Content-Range not specified')
 
         # Get Content-Range and check if Range is consistent with server state
         file_range = ContentRange.from_request()
         if not item.data.size == file_range.begin:
-            return ('Content-Range inconsistent. Last byte on Server: %d' % item.data.size), 409
+            raise Conflict(description='Content-Range inconsistent. Last byte on Server: %d' % item.data.size)
 
         # Decode Base64 encoded request data
         try:
             raw_data = base64.b64decode(request.data)
             file_data = BytesIO(raw_data)
         except (base64.binascii.Error, TypeError):
-            return 'Could not decode data body', 400
+            raise BadRequest(description='Could not decode data body')
 
         # Write data chunk to item
         Upload.data(item, file_data, len(raw_data), file_range.begin)
@@ -125,7 +125,7 @@ class ItemUploadView(RestBase):
         The first check is the provided Content-Length. The second is the actual file size on the server.
         """
         if not may(CREATE):
-            return 'Missing Permissions', 403
+            raise Forbidden()
 
         # Collect all expected data from the Request
         file_type = request.headers.get("Content-Type")
@@ -158,13 +158,13 @@ class ItemUploadView(RestBase):
             item = current_app.storage.openwrite(name)
             new_item = False
 
-        res = None
+        response = None
         try:
-            res = self.update_item(item, name)
-            return res
+            response = self.update_item(item, name)
+            return response
         finally:
             # If error response or exception on a new item path, remove item
-            if new_item and not isinstance(res, Response):
+            if new_item and not isinstance(response, Response):
                 current_app.storage.remove(name)
 
     @rest_errorhandler
@@ -181,7 +181,7 @@ class ItemUploadView(RestBase):
             }
         """
         if not may(LIST):
-            return 'Missing Permissions', 403
+            raise Forbidden()
         ret = {}
         for meta in file_infos():
             name = meta.pop(ID)
--- a/src/bepasty/apis/rest.py
+++ b/src/bepasty/apis/rest.py
@@ -6,7 +6,7 @@ from io import BytesIO
 from flask import Response, make_response, url_for, jsonify, stream_with_context, request, current_app
 from flask.views import MethodView
 
-from ..constants import COMPLETE, FILENAME, SIZE, TYPE, TRANSACTION_ID
+from ..constants import COMPLETE, FILENAME, LOCKED, SIZE, TYPE, TRANSACTION_ID
 from ..utils._compat import bytes_type
 from ..utils.date_funcs import get_maxlife
 from ..utils.http import ContentRange, DownloadRange
@@ -163,7 +163,7 @@ class ItemDownloadView(MethodView):
                 return 'File not found', 404
             raise
 
-        if item.meta.get():
+        if item.meta.get(LOCKED):
             error = 'File Locked.'
         elif not item.meta.get(COMPLETE):
             error = 'Upload incomplete. Try again later.'
--- a/src/bepasty/apis/rest.py
+++ b/src/bepasty/apis/rest.py
@@ -5,6 +5,7 @@ from io import BytesIO
 
 from flask import Response, make_response, url_for, jsonify, stream_with_context, request, current_app
 from flask.views import MethodView
+from werkzeug.exceptions import HTTPException, InternalServerError, MethodNotAllowed
 
 from ..constants import COMPLETE, FILENAME, LOCKED, SIZE, TYPE, TRANSACTION_ID
 from ..utils._compat import bytes_type
@@ -15,7 +16,40 @@ from ..utils.permissions import CREATE, LIST, READ, may
 from ..utils.upload import Upload, background_compute_hash
 
 
-class ItemUploadView(MethodView):
+# This wrappper handles exceptions in the REST api implementation.
+#
+# The @blueprint.add_errorhandler decorator could do this, but we have
+# "/lodgeit/" in the same blueprint and there is no way to exclude
+# from using the same error handler ("/lodgeit/" is not REST api).
+def rest_errorhandler(func):
+    def handler(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except HTTPException as exc:
+            return exc.description, exc.code
+        except Exception:
+            if current_app.propagate_exceptions:
+                # if testing/debug mode, re-raise
+                raise
+            exc = InternalServerError()
+            return exc.description, exc.code
+
+    return handler
+
+
+# Default handlers for REST api to handle error
+class RestBase(MethodView):
+    @rest_errorhandler
+    def get(self, *args, **kwargs):
+        raise MethodNotAllowed()
+
+    @rest_errorhandler
+    def post(self, *args, **kwargs):
+        raise MethodNotAllowed()
+
+
+class ItemUploadView(RestBase):
+    @rest_errorhandler
     def post(self):
         """
         Upload file via REST-API. Chunked Upload is supported.
@@ -117,6 +151,7 @@ class ItemUploadView(MethodView):
 
         return response
 
+    @rest_errorhandler
     def get(self):
         """
         Return the list of all files in bepasty, including metadata in the form:
@@ -139,7 +174,8 @@ class ItemUploadView(MethodView):
         return jsonify(ret)
 
 
-class ItemDetailView(MethodView):
+class ItemDetailView(RestBase):
+    @rest_errorhandler
     def get(self, name):
         if not may(READ):
             return 'Missing Permissions', 403
@@ -149,9 +185,10 @@ class ItemDetailView(MethodView):
                             'file-meta': dict(item.meta)})
 
 
-class ItemDownloadView(MethodView):
+class ItemDownloadView(RestBase):
     content_disposition = 'attachment'
 
+    @rest_errorhandler
     def get(self, name):
         if not may(READ):
             return 'Missing Permissions', 403
@@ -202,7 +239,8 @@ class ItemDownloadView(MethodView):
         return ret
 
 
-class InfoView(MethodView):
+class InfoView(RestBase):
+    @rest_errorhandler
     def get(self):
         return jsonify({'MAX_BODY_SIZE': current_app.config['MAX_BODY_SIZE'],
                         'MAX_ALLOWED_FILE_SIZE': current_app.config['MAX_ALLOWED_FILE_SIZE']})
--- a/src/bepasty/apis/rest.py
+++ b/src/bepasty/apis/rest.py
@@ -50,6 +50,52 @@ class RestBase(MethodView):
 
 
 class ItemUploadView(RestBase):
+    def update_item(self, item, name):
+        # Check the actual size of the file on the server against limit
+        # Either 0 if new file or n bytes of already uploaded file
+        Upload.filter_size(item.data.size)
+
+        # Check Content-Range. Needs to be specified, even if only one chunk
+        if not request.headers.get("Content-Range"):
+            return 'Content-Range not specified', 400
+
+        # Get Content-Range and check if Range is consistent with server state
+        file_range = ContentRange.from_request()
+        if not item.data.size == file_range.begin:
+            return ('Content-Range inconsistent. Last byte on Server: %d' % item.data.size), 409
+
+        # Decode Base64 encoded request data
+        try:
+            raw_data = base64.b64decode(request.data)
+            file_data = BytesIO(raw_data)
+        except (base64.binascii.Error, TypeError):
+            return 'Could not decode data body', 400
+
+        # Write data chunk to item
+        Upload.data(item, file_data, len(raw_data), file_range.begin)
+
+        # Make a Response and create Transaction-ID from ItemName
+        response = make_response()
+        name_b = name if isinstance(name, bytes_type) else name.encode()
+        trans_id_b = base64.b64encode(name_b)
+        trans_id_s = trans_id_b if isinstance(trans_id_b, str) else trans_id_b.decode()
+        response.headers[TRANSACTION_ID] = trans_id_s
+
+        # Check if file is completely uploaded and set meta
+        if file_range.is_complete:
+            Upload.meta_complete(item, '')
+            item.meta[SIZE] = item.data.size
+            item.close()
+            background_compute_hash(current_app.storage, name)
+            # Set status 'successful' and return the new URL for the uploaded file
+            response.status = '201'
+            response.headers["Content-Location"] = url_for('bepasty_apis.items_detail', name=name)
+        else:
+            item.close()
+            response.status = '200'
+
+        return response
+
     @rest_errorhandler
     def post(self):
         """
@@ -102,6 +148,7 @@ class ItemUploadView(RestBase):
             Upload.meta_new(item, 0, file_name, file_type,
                             'application/octet-stream',
                             name, maxlife_stamp=maxlife_timestamp)
+            new_item = True
         else:
             # Get file name from Transaction-ID and open from Storage
             trans_id_s = request.headers.get(TRANSACTION_ID)
@@ -109,48 +156,16 @@ class ItemUploadView(RestBase):
             name_b = base64.b64decode(trans_id_b)
             name = name_b if isinstance(name_b, str) else name_b.decode()
             item = current_app.storage.openwrite(name)
+            new_item = False
 
-        # Check the actual size of the file on the server against limit
-        # Either 0 if new file or n bytes of already uploaded file
-        Upload.filter_size(item.data.size)
-
-        # Check Content-Range. Needs to be specified, even if only one chunk
-        if not request.headers.get("Content-Range"):
-            return 'Content-Range not specified', 400
-
-        # Get Content-Range and check if Range is consistent with server state
-        file_range = ContentRange.from_request()
-        if not item.data.size == file_range.begin:
-            return ('Content-Range inconsistent. Last byte on Server: %d' % item.data.size), 409
-
-        # Decode Base64 encoded request data
-        raw_data = base64.b64decode(request.data)
-        file_data = BytesIO(raw_data)
-
-        # Write data chunk to item
-        Upload.data(item, file_data, len(raw_data), file_range.begin)
-
-        # Make a Response and create Transaction-ID from ItemName
-        response = make_response()
-        name_b = name if isinstance(name, bytes_type) else name.encode()
-        trans_id_b = base64.b64encode(name_b)
-        trans_id_s = trans_id_b if isinstance(trans_id_b, str) else trans_id_b.decode()
-        response.headers[TRANSACTION_ID] = trans_id_s
-
-        # Check if file is completely uploaded and set meta
-        if file_range.is_complete:
-            Upload.meta_complete(item, '')
-            item.meta[SIZE] = item.data.size
-            item.close()
-            background_compute_hash(current_app.storage, name)
-            # Set status 'successful' and return the new URL for the uploaded file
-            response.status = '201'
-            response.headers["Content-Location"] = url_for('bepasty_apis.items_detail', name=name)
-        else:
-            item.close()
-            response.status = '200'
-
-        return response
+        res = None
+        try:
+            res = self.update_item(item, name)
+            return res
+        finally:
+            # If error response or exception on a new item path, remove item
+            if new_item and not isinstance(res, Response):
+                current_app.storage.remove(name)
 
     @rest_errorhandler
     def get(self):
--- a/src/bepasty/apis/rest.py
+++ b/src/bepasty/apis/rest.py
@@ -1,19 +1,20 @@
-import errno
 import base64
 import time
 from io import BytesIO
 
 from flask import Response, make_response, url_for, jsonify, stream_with_context, request, current_app
 from flask.views import MethodView
-from werkzeug.exceptions import HTTPException, InternalServerError, MethodNotAllowed
+from werkzeug.exceptions import HTTPException, Conflict, InternalServerError, MethodNotAllowed
 
-from ..constants import COMPLETE, FILENAME, LOCKED, SIZE, TYPE, TRANSACTION_ID
+from ..constants import FILENAME, ID, SIZE, TYPE, TRANSACTION_ID
 from ..utils._compat import bytes_type
 from ..utils.date_funcs import get_maxlife
 from ..utils.http import ContentRange, DownloadRange
 from ..utils.name import ItemName
-from ..utils.permissions import CREATE, LIST, READ, may
+from ..utils.permissions import CREATE, LIST, may
 from ..utils.upload import Upload, background_compute_hash
+from ..views.filelist import file_infos
+from ..views.download import DownloadView
 
 
 # This wrappper handles exceptions in the REST api implementation.
@@ -167,49 +168,28 @@ class ItemUploadView(RestBase):
         if not may(LIST):
             return 'Missing Permissions', 403
         ret = {}
-        for name in current_app.storage:
-            item = current_app.storage.open(name)
+        for meta in file_infos():
+            name = meta.pop(ID)
             ret[name] = {'uri': url_for('bepasty_apis.items_detail', name=name),
-                         'file-meta': dict(item.meta)}
+                         'file-meta': meta}
         return jsonify(ret)
 
 
-class ItemDetailView(RestBase):
-    @rest_errorhandler
-    def get(self, name):
-        if not may(READ):
-            return 'Missing Permissions', 403
+class ItemDetailView(DownloadView, RestBase):
+    def err_incomplete(self, item, error):
+        raise Conflict(description=error)
 
-        with current_app.storage.open(name) as item:
-            return jsonify({'uri': url_for('bepasty_apis.items_detail', name=name),
-                            'file-meta': dict(item.meta)})
-
-
-class ItemDownloadView(RestBase):
-    content_disposition = 'attachment'
+    def response(self, item, name):
+        return jsonify({'uri': url_for('bepasty_apis.items_detail', name=name),
+                        'file-meta': dict(item.meta)})
 
     @rest_errorhandler
     def get(self, name):
-        if not may(READ):
-            return 'Missing Permissions', 403
+        return super(ItemDetailView, self).get(name)
 
-        try:
-            item = current_app.storage.open(name)
-        except (OSError, IOError) as e:
-            if e.errno == errno.ENOENT:
-                return 'File not found', 404
-            raise
 
-        if item.meta.get(LOCKED):
-            error = 'File Locked.'
-        elif not item.meta.get(COMPLETE):
-            error = 'Upload incomplete. Try again later.'
-        else:
-            error = None
-        if error:
-            item.close()
-            return error, 403
-
+class ItemDownloadView(ItemDetailView):
+    def response(self, item, name):
         request_range = DownloadRange.from_request()
         if not request_range:
             range_end = item.data.size - 1
@@ -221,15 +201,7 @@ class ItemDownloadView(RestBase):
                 range_end = min(request_range.end, item.data.size - 1)
             range_begin = request_range.begin
 
-        def stream(begin, end):
-            offset = max(0, begin)
-            with item as _item:
-                while offset <= end:
-                    buf = _item.data.read(16 * 1024, offset)
-                    offset += len(buf)
-                    yield buf
-
-        ret = Response(stream_with_context(stream(range_begin, range_end)))
+        ret = Response(stream_with_context(self.stream(item, range_begin, range_end + 1)))
         ret.headers['Content-Disposition'] = '{0}; filename="{1}"'.format(
             self.content_disposition, item.meta[FILENAME])
         ret.headers['Content-Length'] = (range_end - range_begin) + 1
@@ -237,6 +209,10 @@ class ItemDownloadView(RestBase):
         ret.status = '200'
         ret.headers['Content-Range'] = ('bytes %d-%d/%d' % (range_begin, range_end, item.data.size))
         return ret
+
+    @rest_errorhandler
+    def get(self, name):
+        return super(ItemDetailView, self).get(name)
 
 
 class InfoView(RestBase):
--- a/src/bepasty/tests/test_rest_server.py
+++ b/src/bepasty/tests/test_rest_server.py
@@ -546,6 +546,20 @@ def test_bad_data(client_fixture):
     assert len(os.listdir(app.config['STORAGE_FILESYSTEM_DIRECTORY'])) == 0
 
 
+def test_upload_maxlife(client_fixture):
+    """upload maxlife"""
+
+    _, client, _ = client_fixture
+
+    lifetime = ['foo', 'MINUTES']
+    response = _upload(client, UPLOAD_DATA, token='full', lifetime=lifetime)
+    assert response.status_code == 400
+
+    lifetime = ['1', 'foo']
+    response = _upload(client, UPLOAD_DATA, token='full', lifetime=lifetime)
+    assert response.status_code == 400
+
+
 def test_upload_too_big(client_fixture):
     app, client, _ = client_fixture
 
--- a/src/bepasty/tests/test_rest_server.py
+++ b/src/bepasty/tests/test_rest_server.py
@@ -469,6 +469,22 @@ def test_upload_range(client_fixture):
     check_err_response(response, 409)
 
 
+def test_bad_data(client_fixture):
+    app, client, _ = client_fixture
+
+    filename = 'test.py'
+    ftype = 'text/x-python'
+
+    # upload invalid base64 encode
+    range_str = 'bytes 0-{}/{}'.format(len(UPLOAD_DATA) - 1, len(UPLOAD_DATA))
+    response = _upload(client, UPLOAD_DATA, token='full', filename=filename,
+                       ftype=ftype, range_str=range_str, encode=False)
+    check_err_response(response, 400)
+
+    # server must not have left garbage files
+    assert len(os.listdir(app.config['STORAGE_FILESYSTEM_DIRECTORY'])) == 0
+
+
 def test_upload_too_big(client_fixture):
     app, client, _ = client_fixture
 
--- a/src/bepasty/tests/test_rest_server.py
+++ b/src/bepasty/tests/test_rest_server.py
@@ -569,6 +569,29 @@ def test_upload_maxlife(client_fixture):
     assert response.status_code == 400
 
 
+def test_upload_transaction_id(client_fixture):
+    app, client, faketime = client_fixture
+
+    faketime.set_time(100)
+
+    sep = 10
+    data = UPLOAD_DATA[sep:]
+    range_str = 'bytes {}-{}/{}'.format(sep, len(UPLOAD_DATA) - 1,
+                                        len(UPLOAD_DATA))
+
+    # upload with invalid Transaction-ID
+    trans_id = 'invalid'
+    response = _upload(client, data, token='full', range_str=range_str,
+                       trans_id=trans_id)
+    check_err_response(response, 400)
+
+    # upload with Transaction-ID for invalid filename
+    trans_id = base64.b64encode(b'invalid')
+    response = _upload(client, data, token='full', range_str=range_str,
+                       trans_id=trans_id)
+    check_err_response(response, 400)
+
+
 def test_upload_too_big(client_fixture):
     app, client, _ = client_fixture
 
--- a/src/bepasty/tests/test_rest_server.py
+++ b/src/bepasty/tests/test_rest_server.py
@@ -8,6 +8,7 @@ import threading
 import time
 import copy
 import hashlib
+import re
 from requests.auth import _basic_auth_str
 from flask import current_app
 from flask import url_for
@@ -145,6 +146,8 @@ def check_response(response, code, ftype='application/json', check_data=True):
 
 def check_err_response(response, code, check_data=True):
     check_response(response, code, 'text/html; charset=utf-8', check_data)
+    # check if doesn't have html tag
+    assert not re.match(r'<.+>', response.data.decode())
 
 
 def check_data_response(response, meta, data, offset=0, total_size=None,
--- a/src/bepasty/tests/test_rest_server.py
+++ b/src/bepasty/tests/test_rest_server.py
@@ -224,7 +224,7 @@ def test_config(client_fixture):
 
 
 def _upload(client, data, token=None, filename=None, ftype=None, lifetime=None,
-            range_str=None, trans_id=None, encode=True):
+            range_str=None, trans_id=None, encode=True, set_range=True):
     if data:
         payload = base64.b64encode(data) if encode else data
         payload_len = len(payload)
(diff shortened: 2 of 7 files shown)


src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,666 @@
...
+    check_response(response, 200, check_data=False)
+
+
+def _upload(client, data, token=None, filename=None, ftype=None, lifetime=None,
+            range_str=None, trans_id=None, no_encode=False):
+    if data:
+        if no_encode:
+            payload = data
+        else:
+            payload = base64.b64encode(data)
```
ThomasWaldmann, line 238: payload = base64.b64encode(data) if encode else data

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,666 @@
...
+    return datas, metas
+
+
+def check_detail_or_download(app, client, fid, meta, data, download=False):
+    url = RestUrl(fid=fid)
+
+    if download:
+        url = url.download
+    else:
+        url = url.detail
```
ThomasWaldmann, line 340: url = url.download if download else url.detail

ThomasWaldmann: (also fix other places)

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,666 @@
...
+                       trans_id=response.headers[TRANSACTION_ID])
+    uri = check_upload_response(response)
+    fid = os.path.basename(uri)
+
+    # check a uploaded item
+    meta = make_meta(UPLOAD_DATA, filename=filename, ftype=ftype, uri=uri)
+    check_detail_or_download(app, client, fid, meta, UPLOAD_DATA)
+    check_detail_or_download(app, client, fid, meta, UPLOAD_DATA, download=True)
+
+    # upload again with not append position
```
ThomasWaldmann, line 538: ?

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,666 @@
...
+        for k, v in tmp_config.items():
+            self.orig_config[k] = self.config[k]
+
+    def __enter__(self):
+        for k, v in self.tmp_config.items():
+            self.config[k] = v
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        for k, v in self.orig_config.items():
+            self.config[k] = self.orig_config[k]
```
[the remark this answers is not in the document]
ThomasWaldmann: using dict() to copy assumes that you do not have a need for `copy.deepcopy`.

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,666 @@
...
+
+
+class RestUrl:
```
ThomasWaldmann, line 99: is this really needed or could you also use flask's `url_for` to build URLs?

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,666 @@
...
+
+    # get incomplete item from list
+    url = RestUrl()
+    headers = add_auth('user', 'full')
+    response = client.get(url.list, headers=headers)
+    check_response(response, 200)
+    assert len(response.json) == 1
+
+    fid = list(response.json.keys())[0]
+    # meta = list(response.json.values())[0]
```
ThomasWaldmann, line 818: ?

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,666 @@
...
+
+def test_upload_range(client_fixture):
+    app, client, faketime = client_fixture
+
+    faketime.set_time(100)
+
+    filename = 'test.py'
+    ftype = 'text/x-python'
+
+    # upload a first half of data
```
ThomasWaldmann: upload first part of data

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,666 @@
...
+    sep = 10
+    data = UPLOAD_DATA[:sep]
+    range_str = 'bytes {}-{}/{}'.format(0, sep - 1, sep + 1)
+    response = _upload(client, data, token='full', filename=filename,
+                       ftype=ftype, range_str=range_str)
+    check_upload_response(response, 200)
+
+    # upload a rest of data
```
ThomasWaldmann, line 524: upload remainder of data

src/bepasty/apis/rest.py
```diff
@@ -15,7 +16,44 @@
 from ..utils.upload import Upload, background_compute_hash
 
 
-class ItemUploadView(MethodView):
+# This wrapper to handle exception of REST api.
+# @blueprint.add_errorhandler decorator can do it though. We have
+# "/lodgeit/" in same blueprint, and didn't find the way to exclude
+# "/lodgeit/" from add_errorhandler to handle exception as web site,
+# not REST api.
```
ThomasWaldmann, line 25: Is this what you mean?
```
This wrappper handles exceptions in the REST api implementation.

The @blueprint.add_errorhandler decorator could do this, but we have
"/lodgeit/" in the same blueprint and there is no way to exclude this from
using the same error handler.
We can't use the same error handler though, because ...
```

src/bepasty/apis/rest.py
```diff
@@ -171,77 +172,52 @@ def get(self):
...
         request_range = DownloadRange.from_request()
         if not request_range:
-            range_end = item.data.size - 1
-            range_begin = 0
+            size = item.data.size
+            start = 0
         else:
             if request_range.end == -1:
-                range_end = item.data.size - 1
+                size = item.data.size
```
ThomasWaldmann, line 227: guess this needs to consider request_range.begin

ThomasWaldmann: in general, if the old way of dealing with begin/end was **not** broken, i'ld prefer you to not change it, but just keep it as is.

src/bepasty/tests/test_rest_server.py
```diff
@@ -477,6 +477,22 @@ def test_upload_range(client_fixture):
...
+    filename = 'test.py'
+    ftype = 'text/x-python'
+
+    # upload invalid base64 encode
+    range_str = 'bytes 0-{}/{}'.format(len(UPLOAD_DATA) - 1, len(UPLOAD_DATA))
+    response = _upload(client, UPLOAD_DATA, token='full', filename=filename,
+                       ftype=ftype, range_str=range_str, no_encode=True)
+    check_err_response(response, 400)
+
+    # server should not left garbage files
```
ThomasWaldmann, line 565: ... must not have left ...

src/bepasty/apis/rest.py
```diff
@@ -54,6 +54,52 @@ def post(self, *args, **kwargs):
...
+        file_range = ContentRange.from_request()
+        if not item.data.size == file_range.begin:
+            return ('Content-Range inconsistent. Last byte on Server: %d' % item.data.size), 409
+
+        # Decode Base64 encoded request data
+        try:
+            raw_data = base64_b64decode(request.data)
+            file_data = BytesIO(raw_data)
+        except (base64.binascii.Error, TypeError):
+            return 'Data body could not decode', 400
```
ThomasWaldmann: 'Could not decode data body'

src/bepasty/tests/test_rest_server.py
```diff
@@ -224,7 +224,7 @@ def test_config(client_fixture):
 
 
 def _upload(client, data, token=None, filename=None, ftype=None, lifetime=None,
-            range_str=None, trans_id=None, no_encode=False):
+            range_str=None, trans_id=None, no_encode=False, no_range=False):
```
ThomasWaldmann, line 233: again: avoid double negation. `range=True`

src/bepasty/tests/test_rest_server.py
```diff
@@ -240,7 +240,8 @@ def _upload(client, data, token=None, filename=None, ftype=None, lifetime=None,
     }
     if range_str is None:
         range_str = 'bytes 0-{}/{}'.format(payload_len - 1, payload_len)
-    headers['Content-Range'] = range_str
+    if not no_range:
```
ThomasWaldmann, line 249: if range: ...

src/bepasty/utils/date_funcs.py
```diff
@@ -10,7 +11,10 @@ def get_maxlife(data, underscore):
     unit = data.get(unit_key, unit_default).upper()
     value_key = 'maxlife_value' if underscore else 'maxlife-value'
     value_default = '1'
-    value = int(data.get(value_key, value_default))
+    try:
+        value = int(data.get(value_key, value_default))
+    except (ValueError, TypeError):
+        raise BadRequest(description='Maxlife-Value header is incorrect')
```
ThomasWaldmann, line 17: guess this should use value_key to generate the description.

src/bepasty/utils/date_funcs.py
```diff
@@ -31,6 +35,8 @@ def time_unit_to_sec(value, unit):
         'YEARS': 60 * 60 * 24 * 365,
         'FOREVER': FOREVER,
     }
+    if units.get(unit) is None:
+        raise BadRequest(description='Maxlife-Unit header is incorrect')
```
ThomasWaldmann, line 39: this function does not really know what it is used for, so it should rather raise a KeyError and let the caller generate the BadRequest.

src/bepasty/utils/upload.py
```diff
@@ -36,9 +37,12 @@ def filter_size(cls, i):
         Filter size.
         Check for advertised size.
         """
-        i = int(i)
+        try:
+            i = int(i)
+        except (ValueError, TypeError):
+            raise BadRequest(description='Size is incorrect')
```
ThomasWaldmann, line 43: Size is invalid

src/bepasty/apis/rest.py
```diff
@@ -157,9 +158,17 @@ def post(self):
             # Get file name from Transaction-ID and open from Storage
             trans_id_s = request.headers.get(TRANSACTION_ID)
             trans_id_b = trans_id_s if isinstance(trans_id_s, bytes_type) else trans_id_s.encode()
-            name_b = base64.b64decode(trans_id_b)
+            try:
+                name_b = base64_b64decode(trans_id_b)
+            except (base64.binascii.Error, TypeError):
+                raise BadRequest(description='{} is incorrect'.format(TRANSACTION_ID))
```
ThomasWaldmann, line 164: do you mean: `could not decode {}` ?

src/bepasty/apis/rest.py
```diff
@@ -157,9 +158,17 @@ def post(self):
...
             name = name_b if isinstance(name_b, str) else name_b.decode()
-            item = current_app.storage.openwrite(name)
+            try:
+                item = current_app.storage.openwrite(name)
+            except (OSError, IOError) as e:
+                if e.errno == errno.ENOENT:
+                    raise BadRequest(description='{} is incorrect'.format(TRANSACTION_ID))
```
ThomasWaldmann, line 170: guess this should be something like "could not find storage item for transaction id"

ThomasWaldmann:
Note: this PR was still too large overall.

I appreciate you putting lots of work into improving bepasty, but please do smaller and more focussed PRs.

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,666 @@
...
+        self.orig = time.time
+        self.now = now
+        self.set_time(self.now)
+
+    def set_time(self, now=None):
+        self.now = now
+        if now and time.time != self.get_time:
+            time.time = self.get_time
+        elif now is None and time.time == self.get_time:
+            time.time = self.orig
```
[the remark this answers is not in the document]
OGAWAHirofumi: faketime.set_time() is used like this too.

```python
        faketime.set_time(3600 * 24 * 365 * 2)
        check_expired(client, fid, download=True)
        # check again (should be deleted already)
        faketime.set_time(100)
```

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,666 @@
...
+    def __enter__(self):
+        self.set_time(self.now)
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.set_time(None)
+
+
+def wait_background():
+    """Wait background threads"""
```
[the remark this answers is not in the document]
OGAWAHirofumi: fixed

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,666 @@
...
+        app.config['TESTING'] = True
+
+        with app.test_client() as client:
+            yield app, client, faketime
+
+        wait_background()
+
+
+class TmpConfig:
+    def __init__(self, app, tmp_config={}):
```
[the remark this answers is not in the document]
OGAWAHirofumi: fixed

src/bepasty/tests/test_rest_server.py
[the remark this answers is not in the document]
OGAWAHirofumi: Hm, we want to backup only part of app.config that specified keys by tmp_config.keys(), so not sure how use dict() or copy.deepcopy.

fixed to use dict.update()

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,666 @@
...
+"""
+
+
+class FakeTime:
+    """Overwrite time.time to control the timestamp that is used in server"""
+
+    def __init__(self, now=None):
+        self.orig = time.time
+        self.now = now
+        self.set_time(self.now)
```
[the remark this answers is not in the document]
OGAWAHirofumi: this is also used as following too, or I'm missing something

```python
        with FakeTime(int(time.time()) + i):
            meta = upload(client, data, token='full', filename=filename,
                          ftype=ftype, lifetime=lifetime)
```

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,666 @@
...
+        return '/apis/rest/items/{}'.format(self.fid)
+
+    @property
+    def download(self):
+        return '/apis/rest/items/{}/download'.format(self.fid)
+
+
+def add_auth(user, password, headers=None):
+    if headers is None:
+        headers = dict()
```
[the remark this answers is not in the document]
OGAWAHirofumi: fixed

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,666 @@
...
+    headers.update({'Authorization': _basic_auth_str(user, password)})
```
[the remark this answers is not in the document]
OGAWAHirofumi: fixed

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,666 @@
...
+    return headers
+
+
+def test_invalid_url(client_fixture):
+    _, client, _ = client_fixture
+
+    # invalid url
+    response = client.get('/apis/rest/foo')
```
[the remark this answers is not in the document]
OGAWAHirofumi: fixed

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,666 @@
...
+def check_err_response(response, code, check_data=True):
+    check_response(response, code, 'text/html; charset=utf-8', check_data)
+
+
+def check_data_response(response, meta, data, offset=None, total_size=None,
+                        check_data=True):
+    ftype = meta['file-meta'][TYPE]
+    filename = meta['file-meta'][FILENAME]
+    if offset is None:
+        offset = 0
```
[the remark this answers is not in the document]
OGAWAHirofumi: fixed

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,666 @@
...
+    if total_size is None:
+        total_size = len(data)
+
+    dispoition = 'attachment; filename="{}"'.format(filename)
```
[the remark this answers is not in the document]
OGAWAHirofumi: fixed

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,666 @@
...
+                                        total_size)
+
+    check_response(response, 200, ftype, check_data)
+    assert response.headers['Content-Disposition'] == dispoition
+    assert response.headers['Content-Range'] == range_str
+    if check_data:
+        assert data == response.data
+
+
+def check_json_response(response, metas, check_data=True):
```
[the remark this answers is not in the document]
OGAWAHirofumi: this can be multiple of meta on list api

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,666 @@
...
+    check_response(response, 200, check_data=check_data)
+    if check_data:
+        assert metas == response.json
+
+
+def check_upload_response(response, code=201, check_data=True):
+    check_response(response, code, 'text/html; charset=utf-8', check_data)
+
+    if code == 200:
+        assert len(response.headers[TRANSACTION_ID]) > 0
```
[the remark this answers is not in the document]
OGAWAHirofumi: this meat, there is Transction-ID, and its length > 0

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,666 @@
...
+
+    # basic auth (unknown user)
+    response = client.get(url.list, headers=add_auth('user', 'invalid'))
+    check_err_response(response, 403)
+    # basic auth (valid user)
+    response = client.get(url.list, headers=add_auth('user', 'full'))
+    check_response(response, 200)
+
+    # token auth (unknown user)
+    response = client.get(url.list + '?token=invalid')
```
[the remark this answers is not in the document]
OGAWAHirofumi: fixed. looks like working with

```python
    with app.app_context():
        with current_app.test_request_context():
            return url_for('bepasty_apis.items_download', name=self.fid)
```

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,666 @@
...
+    assert response.json['MAX_ALLOWED_FILE_SIZE'] == app.config['MAX_ALLOWED_FILE_SIZE']
+    assert response.json['MAX_BODY_SIZE'] == app.config['MAX_BODY_SIZE']
+
+    # get server config (head)
+    response = client.head(url.config)
+    check_response(response, 200, check_data=False)
+
+
+def _upload(client, data, token=None, filename=None, ftype=None, lifetime=None,
+            range_str=None, trans_id=None, no_encode=False):
```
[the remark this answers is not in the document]
OGAWAHirofumi: fixed

src/bepasty/tests/test_rest_server.py
OGAWAHirofumi: fixed

src/bepasty/tests/test_rest_server.py
OGAWAHirofumi: fixed

src/bepasty/tests/test_rest_server.py
OGAWAHirofumi: changed comment

```python
# upload again with item.data.size != start position (invalid position)
```

src/bepasty/tests/test_rest_server.py
OGAWAHirofumi: To get /apis/rest/items, we need this for client context. So I think it is still useful RestUrl for easier and reliable way

```python
        with current_app.test_request_context():
            return url_for('bepasty_apis.items')
```

src/bepasty/tests/test_rest_server.py
OGAWAHirofumi: It was reminder to get meta. removed

src/bepasty/tests/test_rest_server.py
OGAWAHirofumi: fixed

src/bepasty/tests/test_rest_server.py
OGAWAHirofumi: fixed

src/bepasty/apis/rest.py
OGAWAHirofumi: thanks. fixed as following

```python
# This wrappper handles exceptions in the REST api implementation.
#
# The @blueprint.add_errorhandler decorator could do this, but we have
# "/lodgeit/" in the same blueprint and there is no way to exclude
# from using the same error handler ("/lodgeit/" is not REST api).
```

src/bepasty/views/download.py
```diff
@@ -13,58 +13,62 @@
...
+            need_close = True
+            if not item.meta[COMPLETE]:
+                return self.err_incomplete(item, 'Upload incomplete. Try again later.')
+
+            if item.meta[LOCKED] and not may(ADMIN):
+                raise Forbidden()
+
+            if delete_if_lifetime_over(item, name):
+                raise NotFound()
+            need_close = False
```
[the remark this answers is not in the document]
OGAWAHirofumi: yes. Since self.response() uses item.data, item should not be closed.

src/bepasty/apis/rest.py
OGAWAHirofumi: yes, usually I dont change. but this time self.stream() was consolidated and was changed to handle web site.  so -1 and +1 was repeated, and unreadable for me.

anyway, reverted unnecessary changes.

src/bepasty/tests/test_rest_server.py
OGAWAHirofumi: thanks. fixed

src/bepasty/apis/rest.py
OGAWAHirofumi: fixed

src/bepasty/tests/test_rest_server.py
OGAWAHirofumi: fixed, changed to set_range

src/bepasty/tests/test_rest_server.py
OGAWAHirofumi: fixed

src/bepasty/utils/date_funcs.py
OGAWAHirofumi: fixed as following

```python
        field = 'maxlife_value' if underscore else 'Maxlife-Value'
        raise BadRequest(description='{} header is incorrect'.format(field))
```

src/bepasty/utils/date_funcs.py
OGAWAHirofumi: hm, time_unit_to_sec() is only used from get_maxlife() if exclude tests/. it becomes like following. this was your intent?

```python
try:
    get_maxlife(...)
except KeyError:
     raise BadRequest(description='Maxlife-Unit header is incorrect')
```

src/bepasty/utils/upload.py
OGAWAHirofumi: fixed

src/bepasty/apis/rest.py
OGAWAHirofumi: fixed

src/bepasty/apis/rest.py
OGAWAHirofumi: fixed

OGAWAHirofumi:
Hm, then all those patches are depending on previous patches. So if separated more than this, it has to wait the result of previous PR?

OGAWAHirofumi:
this is to show updated patches. If "separated PR and wait result" and "then fix or next PR" is prefered, I will try with new PR (or if there are better way to do it, it would be fine).

src/bepasty/views/download.py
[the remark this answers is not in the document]
ThomasWaldmann: ah, ok.

src/bepasty/tests/test_rest_server.py
ThomasWaldmann: ah, right, better not overwrite useful builtins.

src/bepasty/utils/date_funcs.py
ThomasWaldmann: yes, or use some variable, if that can have underscore alternatively.

src/bepasty/tests/test_rest_server.py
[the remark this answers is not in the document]
ThomasWaldmann: well, if you use a CM, the enter method is called when entering the block inside the CM.

so, faking time can get initialised when entering and normality restored when exiting.

src/bepasty/tests/test_rest_server.py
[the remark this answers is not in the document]
ThomasWaldmann: not sure if doing a partial backup in a python loop is faster than just creating a full backup by copying within the python interpreter. the latter would be easier in any case.

src/bepasty/tests/test_rest_server.py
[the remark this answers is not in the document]
ThomasWaldmann: ah

src/bepasty/tests/test_rest_server.py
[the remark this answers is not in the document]
ThomasWaldmann: ok

src/bepasty/tests/test_rest_server.py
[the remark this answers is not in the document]
ThomasWaldmann: great!

src/bepasty/tests/test_rest_server.py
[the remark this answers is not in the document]
ThomasWaldmann: (also: guess we rather optimize for simplicity here, not for performance)

src/bepasty/tests/test_rest_server.py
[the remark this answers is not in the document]
OGAWAHirofumi: It was for

```python
faketime = FakeTIme(100)
```

usage. but this usage was gone from current code. so removed `.set_time()` call from `__init__`.

src/bepasty/tests/test_rest_server.py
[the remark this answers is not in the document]
OGAWAHirofumi: Hm, I did want to change app.config's dict itself. because flask may have another reference to app.config's dict internally (or future).

e.g.
```python
some_context.config = app.config
...
if some_context.config['FOO']:
    ....
```

so I was not sure if the setting new dict to app.config works or not.

src/bepasty/utils/date_funcs.py
OGAWAHirofumi: ah, ok. changed the get_maxlife() to catch KeyError, and time_unit_to_sec() keep as is (KeyError is raised naturally if no key)

```python
    try:
        return time_unit_to_sec(value, unit)
    except KeyError:
        field = unit_key if underscore else 'Maxlife-Unit'
        raise BadRequest(description='{} header is incorrect'.format(field))
```

OGAWAHirofumi:
changes:

- caa507a: removed `self.set_time()` from `FakeTime().__init__()`
- b1b46ef: changed to check `KeyError` in `get_maxlife()`, and handled `underscore`

src/bepasty/tests/test_rest_server.py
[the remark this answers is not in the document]
OGAWAHirofumi: ah, i noticed no issue now. so changed to

```python
class TmpConfig:
    def __init__(self, app, tmp_config=None):
        self.config = app.config
        self.orig_config = copy.deepcopy(app.config)
        self.tmp_config = {} if tmp_config is None else tmp_config

    def __enter__(self):
        self.config.update(self.tmp_config)

    def __exit__(self, exc_type, exc_value, traceback):
        self.config.update(self.orig_config)
```

OGAWAHirofumi:
changes:
    
- ae36009: use copy.deepcopy() in TmpConfig

(included in previous update)
- caa507a: removed self.set_time() from FakeTime().__init__()
- b1b46ef: changed to check KeyError in get_maxlife(), and handled underscore

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,817 @@
...
+    def __init__(self, app, tmp_config=None):
+        self.config = app.config
+        self.orig_config = copy.deepcopy(app.config)
+        self.tmp_config = {} if tmp_config is None else tmp_config
+
+    def __enter__(self):
+        self.config.update(self.tmp_config)
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.config.update(self.orig_config)
```
ThomasWaldmann, line 95: do we need self.config? guess directly using app.config is less confusing:

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,817 @@
...
+
+
+class RestUrl:
+    def __init__(self, fid=None):
+        self.fid = fid
```
ThomasWaldmann, line 100: minor nitpick: `fid` is not a very speaking name. bepasty usually talks about items, so how about `item_id`?

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,817 @@
...
+            'none': '',
+        }
+        Config.STORAGE_FILESYSTEM_DIRECTORY = str(tmp_path)
+
+        app = create_app()
+        app.config['TESTING'] = True
+
+        with app.test_client() as client:
+            with app.app_context():
+                yield app, client, faketime
```
ThomasWaldmann, line 80: hmm, would it be easier to have separate fixtures for these?

- app
- client
- request
- faketime

src/bepasty/tests/test_rest_server.py
ThomasWaldmann: if the tests needing a request context would "require" it via a fixture param, guess this RestUrl class could be removed and we could directly use url_for() inside the test function.

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,817 @@
...
+
+
+def check_response(response, code, ftype='application/json', check_data=True):
+    assert response.status_code == code
+    assert response.headers['Content-Type'] == ftype
+    if check_data:
+        assert int(response.headers['Content-Length']) == len(response.data)
+
+
+re_html_tag = re.compile(r'<.+>')
```
ThomasWaldmann, line 147: if this is only used in the function below, move it into that scope. we don't need to micro-optimize here.

src/bepasty/utils/date_funcs.py
```diff
@@ -1,17 +1,26 @@
...
     unit = data.get(unit_key, unit_default).upper()
     value_key = 'maxlife_value' if underscore else 'maxlife-value'
     value_default = '1'
-    value = int(data.get(value_key, value_default))
-    return time_unit_to_sec(value, unit)
+    try:
+        value = int(data.get(value_key, value_default))
+    except (ValueError, TypeError):
+        field = value_key if underscore else 'Maxlife-Value'
+        raise BadRequest(description='{} header is incorrect'.format(field))
```
ThomasWaldmann, line 18: remove 17 and just use value_key instead of field?

src/bepasty/utils/date_funcs.py
```diff
@@ -1,17 +1,26 @@
...
+    try:
+        return time_unit_to_sec(value, unit)
+    except KeyError:
+        field = unit_key if underscore else 'Maxlife-Unit'
+        raise BadRequest(description='{} header is incorrect'.format(field))
```
ThomasWaldmann, line 23: remove 22 and just use unit_key instead of field?

src/bepasty/tests/test_rest_server.py
OGAWAHirofumi: this test_request_context() is **fake** request context, just to make url_for() works. so I don't want put other code under this context.

src/bepasty/tests/test_rest_server.py
OGAWAHirofumi: hm, then we have to make app global? IMO, self.config looks better.

src/bepasty/tests/test_rest_server.py
OGAWAHirofumi: easier would be one. because no need to think/handle other combinations.

src/bepasty/utils/date_funcs.py
OGAWAHirofumi: other header fields uses camel case. if you don't care inconsistency, value_key would be fine.

src/bepasty/utils/date_funcs.py
OGAWAHirofumi: same with value_key

src/bepasty/tests/test_rest_server.py
OGAWAHirofumi: i wanted to use shorter name though. changed to item_id

src/bepasty/tests/test_rest_server.py
OGAWAHirofumi: changed

OGAWAHirofumi:
changes:
- 61c3e3d: rename fid to item_id
- c2b5f34: use re.match simply

(included in previous update)
- ae36009: use `copy.deepcopy()` in `TmpConfig`
- caa507a: removed `self.set_time()` from `FakeTime().__init__()`
- b1b46ef: changed to check `KeyError` in `get_maxlife()`, and handled `underscore`

OGAWAHirofumi:
ping?

src/bepasty/utils/_compat.py
```diff
@@ -14,9 +15,15 @@
...
+        return base64.b64decode(data)
 else:
     from urllib.parse import urlparse, urljoin
 
     range_type = range
     bytes_type = bytes
     iteritems = lambda d: iter(d.items())  # noqa: E731
+
+    def base64_b64decode(data):
+        return base64.b64decode(data, validate=True)
```
ThomasWaldmann, line 29: Maybe this is well-meant (like "validating is always better than not"), but in the end, this implements different behaviour for python 2.7 and python 3.x. I don't think we want different behaviour.

So, there should be a ticket about changing this to validating **after** we dropy python 2 support, but this should not be changed within this PR.

OGAWAHirofumi: Hm, maybe no. in my understand, python 2.7 doesn't have validate parameter and validate by default. but python3 default is no validation.

So this sets validate to make be same behavior.

ThomasWaldmann: no, 2.7 does not validate by default, it strips invalid stuff, like default in 3.x - see the docs.

OGAWAHirofumi: - https://docs.python.org/2.7/library/base64.html
The decoded string is returned. A TypeError is raised if s is incorrectly padded. Characters that are neither in the normal base-64 alphabet nor the alternative alphabet are discarded prior to the padding check.

- https://docs.python.org/3/library/base64.html
If validate is False (the default), characters that are neither in the normal base-64 alphabet nor the alternative alphabet are discarded prior to the padding check. If validate is True, these non-alphabet characters in the input result in a binascii.Error.

Looks like it rasise TypeError?

ThomasWaldmann: characters being discarded is the common default, so both behave the same.

OGAWAHirofumi: hm, sure. looks like I somehow confused when testing.

OGAWAHirofumi:
changes:
- removed compat base64_b64decode completely

(included in previous update)
- 61c3e3d: rename fid to item_id
- c2b5f34: use re.match simply
- ae36009: use `copy.deepcopy()` in `TmpConfig`
- caa507a: removed `self.set_time()` from `FakeTime().__init__()`
- b1b46ef: changed to check `KeyError` in `get_maxlife()`, and handled `underscore`

src/bepasty/utils/date_funcs.py
ThomasWaldmann: i think 17 is just inconsistent.

we access via value_key, always.

so the error message should not use value_key for one case and something hardcoded for the other.

src/bepasty/utils/date_funcs.py
ThomasWaldmann: same here.

src/bepasty/apis/rest.py
```diff
@@ -5,17 +5,103 @@
...
-class ItemUploadView(MethodView):
+# This wrappper handles exceptions in the REST api implementation.
+#
+# The @blueprint.add_errorhandler decorator could do this, but we have
+# "/lodgeit/" in the same blueprint and there is no way to exclude
+# from using the same error handler ("/lodgeit/" is not REST api).
+def rest_errorhandler(func):
+    def error_message(description, code):
+        # maybe better to return error as json?
+        return description, code
```
ThomasWaldmann, line 29: remove this function

src/bepasty/apis/rest.py
```diff
@@ -5,17 +5,103 @@
...
+
+    def handler(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except HTTPException as e:
+            return error_message(e.description, e.code)
```
ThomasWaldmann, line 35:
```
... as exc:
    pass
```

src/bepasty/apis/rest.py
```diff
@@ -5,17 +5,103 @@
...
+        except Exception:
+            if current_app.propagate_exceptions:
+                # if testing/debug mode, re-raise
+                raise
+            exc = InternalServerError()
+            return error_message(exc.description, exc.code)
```
ThomasWaldmann, line 41: outdent this line 1 level and use:
```
return exc.description, exc.code
```

OGAWAHirofumi: you meant the following?

```python
def rest_errorhandler(func):
    def handler(*args, **kwargs):
        try:
            return func(*args, **kwargs)
        except HTTPException as exc:
            pass
        except Exception:
            if current_app.propagate_exceptions:
                # if testing/debug mode, re-raise
                raise
            exc = InternalServerError()
        return exc.description, exc.code

    return handler
```

if so, looks like not working,
```
src/bepasty/apis/rest.py:28:9: F841 local variable 'exc' is assigned to but never used
```

src/bepasty/utils/date_funcs.py
OGAWAHirofumi: users of get_maxlife() are, one is request.form and one is request.headers.

And questionable one is request.headers, because it is case insensitive dict, and the header is usually using camel case. If we don't care this camel case in request.headers,

for example,

```python
        if not request.headers.get("Content-Range"):
            return 'Content-Range not specified', 400
```

Content-Range error uses camel case. So actually I myself don't care lower or camel though, I wanted to confirm you also don't care this.

src/bepasty/utils/date_funcs.py
ThomasWaldmann: ok

src/bepasty/utils/date_funcs.py
ThomasWaldmann: don't care, we'll just use the key (which we really use to access the dict) also for the message.

src/bepasty/apis/rest.py
ThomasWaldmann: ehrm. either i do not know some strange scoping rule or this is a false positive from flake8 tester...

src/bepasty/utils/date_funcs.py
OGAWAHirofumi: the idea to remove "field = ..." stuff, how about the following change? using camel case for *_key

```python
def get_maxlife(data, underscore):
    unit_key = 'maxlife_unit' if underscore else 'Maxlife-Unit'
    unit_default = 'MONTHS'
    unit = data.get(unit_key, unit_default).upper()
    value_key = 'maxlife_value' if underscore else 'Maxlife-Value'
    value_default = '1'
    try:
        value = int(data.get(value_key, value_default))
    except (ValueError, TypeError):
        raise BadRequest(description='{} header is incorrect'.format(value_key))
    try:
        return time_unit_to_sec(value, unit)
    except KeyError:
        raise BadRequest(description='{} header is incorrect'.format(unit_key))
```

src/bepasty/apis/rest.py
OGAWAHirofumi: it was actual error in python3 at least

ThomasWaldmann: ok, looks like the first. exc is only defined inside the except block.

```
def rest_errorhandler(func):
    def handler(*args, **kwargs):
        try:
            return func(*args, **kwargs)
        except HTTPException as exc:
            return exc.description, exc.code
        except Exception:
            if current_app.propagate_exceptions:
                # if testing/debug mode, re-raise
                raise
            exc = InternalServerError()
            return exc.description, exc.code

    return handler
```

OGAWAHirofumi:
changes:
- removed error_message() from rest_errorhandler
- removed "field = ..." and use camel case for get_maxlife()

(included in previous update)
- removed compat base64_b64decode completely
- 61c3e3d: rename fid to item_id
- c2b5f34: use re.match simply
- ae36009: use `copy.deepcopy()` in `TmpConfig`
- caa507a: removed `self.set_time()` from `FakeTime().__init__()`
- b1b46ef: changed to check `KeyError` in `get_maxlife()`, and handled `underscore`

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,816 @@
...
+#
+
+import os
+import base64
+import threading
+import time
+import copy
+import hashlib
+import re
+from requests.auth import _basic_auth_str
```
ThomasWaldmann, line 12: this does not look like a public function, can you search for a better way?

src/bepasty/utils/date_funcs.py
```diff
@@ -1,17 +1,24 @@
...
 
 
 def get_maxlife(data, underscore):
-    unit_key = 'maxlife_unit' if underscore else 'maxlife-unit'
-    unit_default = 'MONTH'
+    unit_key = 'maxlife_unit' if underscore else 'Maxlife-Unit'
+    unit_default = 'MONTHS'
     unit = data.get(unit_key, unit_default).upper()
-    value_key = 'maxlife_value' if underscore else 'maxlife-value'
+    value_key = 'maxlife_value' if underscore else 'Maxlife-Value'
```
ThomasWaldmann, line 12: if maxlife-unit and maxlife-value were not broken, do not "fix" them.

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,816 @@
...
+                       ftype=ftype, range_str=range_str)
+    check_err_response(response, 400)
+
+    # upload first part of data
+    range_str = 'bytes {}-{}/{}'.format(0, sep - 1, len(UPLOAD_DATA))
+    response = _upload(client, data, token='full', filename=filename,
+                       ftype=ftype, range_str=range_str)
+    check_upload_response(response, 200)
+
+    # upload reminder of data
```
ThomasWaldmann, line 512: remainder (please fix globally)

src/bepasty/tests/test_rest_server.py
```diff
@@ -0,0 +1,816 @@
...
+    data = UPLOAD_DATA[sep:]
+    range_str = 'bytes {}-{}/{}'.format(sep, len(UPLOAD_DATA) - 1,
```
ThomasWaldmann, line 514: shouldn't this be: `sep, sep + len(UPLOAD_DATA) - 1, len(UPLOAD_DATA)`

and if so, why did the test succeed? bug?

ThomasWaldmann: ah, no, my fault. it would be sep + len(data) - 1, but your version is also correct.

src/bepasty/utils/date_funcs.py
OGAWAHirofumi: ok, if you think so. removed

src/bepasty/tests/test_rest_server.py
OGAWAHirofumi: fixed. (there was no others)

src/bepasty/tests/test_rest_server.py
OGAWAHirofumi: I cant guess your favorite though, how about this?

```python
def add_auth(user, password, headers=None):
    req = Request('POST', 'http://dummy', auth=(user, password))

    headers = headers if headers is not None else {}
    headers['Authorization'] = req.prepare().headers['Authorization']
    return headers
```

OGAWAHirofumi:
changes:
- removed Maxlife-*
- fix typo reminder => remainder
- try other way for basic auth

(included in previous update)
- removed error_message() from rest_errorhandler
- removed "field = ..." and use camel case for get_maxlife()
- removed compat base64_b64decode completely
- 61c3e3d: rename fid to item_id
- c2b5f34: use re.match simply
- ae36009: use `copy.deepcopy()` in `TmpConfig`
- caa507a: removed `self.set_time()` from `FakeTime().__init__()`
- b1b46ef: changed to check `KeyError` in `get_maxlife()`, and handled `underscore`

src/bepasty/tests/test_rest_server.py
ThomasWaldmann: constructing a full Request ovjects seems a bit much, I thought there should be some public function somewhere that does what we need.

but functions starting with single or double underscore are usually not meant to be a public stable API, so that usually means "don't use this from outside".

ThomasWaldmann: i looked around and found it rather embarrassing to not being able to find a reusable implementation of that in python stdlib, werkzeug, requests. everywhere it is just hardcoded inside some other stuff.

so, ignore my feedback and just use that internal function until we find something better.

src/bepasty/utils/date_funcs.py
ThomasWaldmann: i guess within the current commits in this PR, you have the least possible changes, so keep it as is.

in general, i do not like huge pull requests and i especially do not like even more huge pull requests that additionally "fix" (change) stuff that was **not** broken.

the work you do within this PR is great and very welcome (esp. that you added a lot of tests), but let's concentrate on real fixes here.

OGAWAHirofumi: I agree, but at same time disagree. making clean history is good and very preferred, however, it takes more time and cost. so we would need a balance as usual. and IMO, at least, making forward and better result should be more important than cleaner history.

anyway, i already reverted those changes.

src/bepasty/tests/test_rest_server.py
OGAWAHirofumi: ok, i will make PR with _basic_auth_str later.

ThomasWaldmann:
Note: added this to 0.6.0 (next release) milestone.

OGAWAHirofumi:
changes:
- reverted "try other way for basic auth", now use _basic_auth_str()

(included in previous update)
- removed Maxlife-*
- fix typo reminder => remainder
- removed error_message() from rest_errorhandler
- removed "field = ..." and use camel case for get_maxlife()
- removed compat base64_b64decode completely
- 61c3e3d: rename fid to item_id
- c2b5f34: use re.match simply
- ae36009: use `copy.deepcopy()` in `TmpConfig`
- caa507a: removed `self.set_time()` from `FakeTime().__init__()`
- b1b46ef: changed to check `KeyError` in `get_maxlife()`, and handled `underscore`

src/bepasty/tests/test_rest_server.py
ThomasWaldmann: thanks!

ThomasWaldmann:
Yay, merged! Thanks for your PR!

PR #217, opened 2020-10-17
Support reverse proxy setup

OGAWAHirofumi:
On reverse proxy setup, some uses wrong URL, for example, QR uses
internal hostname.

To fixes, this adds new "PUBLIC_URL" in config to tell URL for
frontend side, then uses it to setup flask environ to use proper URL.

### DIFF
--- a/src/bepasty/app.py
+++ b/src/bepasty/app.py
@@ -10,6 +10,7 @@ from flask import (
     render_template,
     session,
 )
+from werkzeug.middleware.proxy_fix import ProxyFix
 
 from .apis import blueprint as blueprint_apis
 from .storage import create_storage
@@ -55,14 +56,13 @@ class ReverseProxyMiddleware(PrefixMiddleware):
             super(ReverseProxyMiddleware, self).__init__(app, prefix=prefix)
 
     def __call__(self, environ, start_response):
-        scheme = environ.get('HTTP_X_FORWARDED_PROTO') or self.scheme
-        if scheme:
-            environ['wsgi.url_scheme'] = scheme
-        host = environ.get('HTTP_X_FORWARDED_HOST') or self.server
-        if host:
-            environ['HTTP_HOST'] = host
-        script_name = environ.get('HTTP_X_FORWARDED_PREFIX') or self.script_name
-        environ['SCRIPT_NAME'] = script_name
+        # setup environ if not setup by ProxyFix()
+        if not environ.get('HTTP_X_FORWARDED_PROTO'):
+            environ['wsgi.url_scheme'] = self.scheme
+        if not environ.get('HTTP_X_FORWARDED_HOST'):
+            environ['HTTP_HOST'] = self.server
+        if not environ.get('HTTP_X_FORWARDED_PREFIX'):
+            environ['SCRIPT_NAME'] = self.script_name
 
         if self.prefix is not None:
             return super(ReverseProxyMiddleware, self).__call__(
@@ -91,6 +91,10 @@ def create_app():
                                                   prefix=prefix)
         else:
             raise Exception("could not parse PUBLIC_URL [%s]" % public_url)
+
+        # For now, x_for==1. I.e. assuming not nested proxies.
+        app.wsgi_app = ProxyFix(app.wsgi_app, x_for=1, x_proto=1,
+                                x_host=1, x_port=1, x_prefix=1)
     elif prefix is not None:
         app.wsgi_app = PrefixMiddleware(app.wsgi_app, prefix=prefix)
 
--- a/src/bepasty/app.py
+++ b/src/bepasty/app.py
@@ -1,5 +1,6 @@
 import os
 import time
+import re
 
 from flask import (
     Flask,
@@ -32,14 +33,43 @@ class PrefixMiddleware(object):
         self.app = app
         self.prefix = prefix
 
-    def __call__(self, environ, start_response):
+    def __call__(self, environ, start_response, set_script_name=True):
         if environ['PATH_INFO'].startswith(self.prefix):
             environ['PATH_INFO'] = environ['PATH_INFO'][len(self.prefix):]
-            environ['SCRIPT_NAME'] = self.prefix
+            if set_script_name:
+                environ['SCRIPT_NAME'] = self.prefix
             return self.app(environ, start_response)
         else:
             start_response('404', [('Content-Type', 'text/plain')])
             return ['This URL does not belong to the bepasty app.'.encode()]
+
+
+class ReverseProxyMiddleware(PrefixMiddleware):
+    def __init__(self, app, scheme, server, script_name='', prefix=None):
+        self.app = app
+        self.scheme = scheme
+        self.server = server
+        self.script_name = script_name or ''
+        self.prefix = prefix
+        if prefix is not None:
+            super(ReverseProxyMiddleware, self).__init__(app, prefix=prefix)
+
+    def __call__(self, environ, start_response):
+        scheme = environ.get('HTTP_X_FORWARDED_PROTO') or self.scheme
+        if scheme:
+            environ['wsgi.url_scheme'] = scheme
+        host = environ.get('HTTP_X_FORWARDED_HOST') or self.server
+        if host:
+            environ['HTTP_HOST'] = host
+        script_name = environ.get('HTTP_X_FORWARDED_PREFIX') or self.script_name
+        environ['SCRIPT_NAME'] = script_name
+
+        if self.prefix is not None:
+            return super(ReverseProxyMiddleware, self).__call__(
+                environ, start_response, set_script_name=False
+            )
+        else:
+            return self.app(environ, start_response)
 
 
 def create_app():
@@ -49,8 +79,19 @@ def create_app():
     if os.environ.get('BEPASTY_CONFIG'):
         app.config.from_envvar('BEPASTY_CONFIG')
 
+    public_url = app.config.get('PUBLIC_URL')
     prefix = app.config.get('APP_BASE_PATH')
-    if prefix is not None:
+    if public_url is not None:
+        m = re.match('(https?)://([^/]+)(/.*)?', public_url.rstrip('/'))
+        if m:
+            app.wsgi_app = ReverseProxyMiddleware(app.wsgi_app,
+                                                  scheme=m.group(1),
+                                                  server=m.group(2),
+                                                  script_name=m.group(3),
+                                                  prefix=prefix)
+        else:
+            raise Exception("could not parse PUBLIC_URL [%s]" % public_url)
+    elif prefix is not None:
         app.wsgi_app = PrefixMiddleware(app.wsgi_app, prefix=prefix)
 
     app.storage = create_storage(app)
--- a/src/bepasty/config.py
+++ b/src/bepasty/config.py
@@ -20,7 +20,8 @@ class Config(object):
     #: Enable reverse proxy setup. This must set the public URL,
     #: i.e. the URL as seen by user's browser.
     #: Also it may use values set on proxy server,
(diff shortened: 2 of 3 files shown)


OGAWAHirofumi:
ping?

src/bepasty/app.py
```diff
@@ -32,25 +33,63 @@ def __init__(self, app, prefix=''):
...
-            environ['SCRIPT_NAME'] = self.prefix
+            if set_script_name:
+                environ['SCRIPT_NAME'] = self.prefix
             return self.app(environ, start_response)
         else:
             start_response('404', [('Content-Type', 'text/plain')])
             return ['This URL does not belong to the bepasty app.'.encode()]
 
 
+class ReverseProxied(PrefixMiddleware):
```
ThomasWaldmann, line 47: `ReverseProxyMiddleware` (follow same naming style)

src/bepasty/app.py
```diff
@@ -32,25 +33,63 @@ def __init__(self, app, prefix=''):
...
+        self.app = app
+        self.scheme = scheme
+        self.server = server
+        self.script_name = script_name or ''
+        self.prefix = prefix
+        if prefix is not None:
+            super(ReverseProxied, self).__init__(app, prefix=prefix)
+
+    def __call__(self, environ, start_response):
+        scheme = environ.get('HTTP_X_FORWARDED_PROTO') or self.scheme
```
ThomasWaldmann, line 58: rather use the default param of get:
```
environ.get(NAME, default)
```

src/bepasty/app.py
```diff
@@ -32,25 +33,63 @@ def __init__(self, app, prefix=''):
...
+        if scheme:
+            environ['wsgi.url_scheme'] = scheme
+        host = environ.get('HTTP_X_FORWARDED_HOST') or self.server
```
ThomasWaldmann, line 61: rather use the default param of get:
```
environ.get(NAME, default)
```

src/bepasty/app.py
```diff
@@ -32,25 +33,63 @@ def __init__(self, app, prefix=''):
...
+        if host:
+            environ['HTTP_HOST'] = host
+        script_name = environ.get('HTTP_X_FORWARDED_PREFIX') or self.script_name
```
ThomasWaldmann, line 64: rather use the default param of get:
```
environ.get(NAME, default)
```

src/bepasty/app.py
```diff
@@ -32,25 +33,63 @@ def __init__(self, app, prefix=''):
...
 
     app.config.from_object('bepasty.config.Config')
     if os.environ.get('BEPASTY_CONFIG'):
         app.config.from_envvar('BEPASTY_CONFIG')
 
+    public_url = app.config.get('PUBLIC_URL')
     prefix = app.config.get('APP_BASE_PATH')
-    if prefix is not None:
+    if public_url is not None:
+        public_url.rstrip('/')
```
ThomasWaldmann, line 85: that does not work. guess you meant:
```
public_url = public_url.rstrip('/')
```

src/bepasty/app.py
```diff
@@ -32,25 +33,63 @@ def __init__(self, app, prefix=''):
...
+        m = re.match('(https?)://([^/]+)(/.*)?', public_url)
+        if m:
+            app.wsgi_app = ReverseProxied(app.wsgi_app, scheme=m.group(1),
+                                          server=m.group(2),
+                                          script_name=m.group(3),
+                                          prefix=prefix)
```
ThomasWaldmann, line 91: what if it does not match?

ThomasWaldmann (commented):
some feedback.

before working on the feedback, can you please check the `ProxyFix` middleware from requests. that is a well-known and widely used middleware.

please test first if ProxyFix does the job and only if not, continue with this PR.

https://werkzeug.palletsprojects.com/en/1.0.x/middleware/proxy_fix/

src/bepasty/app.py
OGAWAHirofumi: I already checked ProxyFix() while working on this. And main reason why I didn't use it is, I couldn't find way to integrate with PrefixMiddleware. And another reason is, changing unnecessary environment for this.

src/bepasty/app.py
OGAWAHirofumi: current behavior - the config is ignored as user error. I'm not sure what we want to do

src/bepasty/app.py
OGAWAHirofumi: fixed

src/bepasty/app.py
OGAWAHirofumi: fixed

src/bepasty/app.py
OGAWAHirofumi: fixed

src/bepasty/app.py
OGAWAHirofumi: ouch, fixed as

```python
        m = re.match('(https?)://([^/]+)(/.*)?', public_url.rstrip('/'))
```

src/bepasty/app.py
ThomasWaldmann: Theoretically, middlewares should be stackable. Why did this not work?

Also, can you rephrase "And another reason is, changing unnecessary environment for this.", I did not understand what you mean.

src/bepasty/app.py
ThomasWaldmann: Not sure if it is the best, but one could `raise Exception("could not parse ENV_VAR_NAME [%s]" % public_url)`.

src/bepasty/app.py
OGAWAHirofumi: > Theoretically, middlewares should be stackable. Why did this not work?

both of PrefixMiddleware() and ProxyFix() modify SCRIPT_NAME, so those are conflict.

> Also, can you rephrase "And another reason is, changing unnecessary environment for this.", I did not understand what you mean.

ProxyFix() modifies REMOTE_ADDR, SERVER_NAME, and SERVER_PORT that we don't use those for now. and especially, to setup REMOTE_ADDR from X-Forwarded-For, we have to decide which part is trustable (hard coding or additional config are required).

src/bepasty/app.py
ThomasWaldmann: why should it be the empty string? either it is set to something that makes sense or it is not set.

src/bepasty/app.py
OGAWAHirofumi: looks like better

src/bepasty/app.py
ThomasWaldmann: modifying SCRIPT_NAME by PrefixMiddleware is solved by your small change in there, right?

what I want to avoid is to invent yet another ProxyFixingMiddleware if werkzeug already has one.

that is code we need to maintain, it might contain unknown bugs. i'ld rather like to leave that to werkzeug, which likely has thousandfold more users than bepasty.

it is not a problem if ProxyFix does some stuff we do not use (yet).
also, iirc, it has a mechanism to only trust trustable headers.

src/bepasty/app.py
ThomasWaldmann: yeah, just replace ENV_VAR_NAME by the real thing.

ThomasWaldmann:
BTW, I run bepasty behind a reverse proxy and I do not see any malfunction (using latest code from master).

src/bepasty/app.py
OGAWAHirofumi: > modifying SCRIPT_NAME by PrefixMiddleware is solved by your small change in there, right?

hm, not sure. I will try again

> what I want to avoid is to invent yet another ProxyFixingMiddleware if werkzeug already has one.
> 
> that is code we need to maintain, it might contain unknown bugs. i'ld rather like to leave that to werkzeug, which likely has thousandfold more users than bepasty.

yes, then failed and copied to own one

> it is not a problem if ProxyFix does some stuff we do not use (yet).
> also, iirc, it has a mechanism to only trust trustable headers.

X-Forwarded-For is not trustable header. it is recording the path of proxies, and client can  behave as proxy to add fake X-Forwarded-For. So it is why ProxyFix() needs trustable number of ip in X-Forwarded-For.

src/bepasty/app.py
OGAWAHirofumi: those are controlled by outside of us. and it was what ProxyFix does

OGAWAHirofumi:
> BTW, I run bepasty behind a reverse proxy and I do not see any malfunction (using latest code from master).
>
> So, I wonder whether this PR is needed at all. Guess in any case there should be first a bug report, precisely describing the problem and then the fix, if it can not be solved outside bepasty (e.g. by setting the usual env vars in the web server).

Tried QR code?

src/bepasty/app.py
ThomasWaldmann: yes, that is known and werkzeug's `ProxyFix` has parameters to specify the number of trusted entries for all the headers it is processing.

i still do not see the reason why we need any other code than that. esp. considering i did not need any code to have bepasty running behind a proxy, not even ProxyFix.

ThomasWaldmann:
Yes, i tried the QR view, looks all like desired there.

ThomasWaldmann:
I didn't mean to use werkzeug ProxyFix inside your additional malware, but to only use ProxyFix instead your new middleware.

Also, this is user specific, so you do not need to add ProxyFix to our code, it can be imported in a custom mybepasty.wsgi script.

OGAWAHirofumi:
> Yes, i tried the QR view, looks all like desired there.
> 
> https://paste.thinkmo.de/3ijPGXTR

I guess, you are running frontend and bepasty are same server?

> I didn't mean to use werkzeug ProxyFix inside your additional malware, but to only use ProxyFix instead your new middleware.

Like said, SCRIPT_NAME is overwritten. So, complex setup becomes impossible.

For example,

public url == https://example.org/apps
internal url == http://localhost/bepasty
 
> Also, this is user specific, so you do not need to add ProxyFix to our code, it can be imported in a custom mybepasty.wsgi script.

you mean, user has to developer and own script for each?

ThomasWaldmann:
I run a frontend nginx for TLS in one VM and another nginx with bepasty in another VM.

You can easily just have the same script name (easy) or tweak this in the web server config.

That stuff is complex, there are all sorts of different bugs / behaviours / brokenness in different web servers and I do not want to maintain that sort of code.

src/bepasty/app.py
```diff
@@ -32,25 +34,68 @@ def __init__(self, app, prefix=''):
...
+    def __call__(self, environ, start_response, set_script_name=True):
         if environ['PATH_INFO'].startswith(self.prefix):
             environ['PATH_INFO'] = environ['PATH_INFO'][len(self.prefix):]
-            environ['SCRIPT_NAME'] = self.prefix
+            if set_script_name:
+                environ['SCRIPT_NAME'] = self.prefix
             return self.app(environ, start_response)
         else:
             start_response('404', [('Content-Type', 'text/plain')])
             return ['This URL does not belong to the bepasty app.'.encode()]
```
ThomasWaldmann, line 45: I guess this is all what we need from this PR and the rest should be done with / within werkzeug.

ThomasWaldmann: And, this behaviour needs to be configured in the middleware `__init__`, not in `__call__`.

OGAWAHirofumi:
> I run a frontend nginx for TLS in one VM and another nginx with bepasty in another VM.
> 
> You can just have the same script name (easy) or tweak this in the web server config.
> 
> That stuff is complex, there are all sorts of different bugs / behaviours / brokenness in different web servers and I do not want to maintain that sort of code. If you think it is important, file a feature request or PR against werkzeug please, not bepasty.

ok. I understood there is no support by bepasty.

PR #218, opened 2020-10-17
Use simple link in list/display view instead of <form>

OGAWAHirofumi:
For example, user can't copy link for "display" button in list
view. To allow to copy link, this changes <form> for GET to simple
link.

### DIFF
--- a/src/bepasty/templates/_utils.html
+++ b/src/bepasty/templates/_utils.html
@@ -42,21 +42,15 @@
                 </td>
                 <td>
                     <div class="btn-group">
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
I also was annoyed already by not being able to use these actions like normal links, so this seems to be a good fix! :-)

Did you practically test this? Does it work? Does it look ok?

OGAWAHirofumi:
I checked by firefox 81, firefox 81 for mobile, and chrome 86, and it looks like no difference with form exception working context menu.

ThomasWaldmann:
Cool! Thanks!

PR #219, opened 2020-10-25
Use bootstrap for QR page

OGAWAHirofumi:
To consistent with other pages, this change QR page to use bootstrap,
and use a same template with display view.

With this change, QR page now has a navigation bar at top like other
pages, and uses better layout, by bootstrap, for several screen size.

[This might be better to use "modal" component of bootstrap in display
view.]

The following is screenshot of result of this change,

qr-page


ThomasWaldmann:
Thanks!

PR #220, opened 2020-10-25
Fix unit_default in date_funcs

mdellweg:


ThomasWaldmann:
Thanks for the PR!

Good catch, but this is already kind-of-fixed in this PR and thus your PR is duplicate:

https://github.com/bepasty/bepasty-server/pull/216/commits/6b4462e47af560339871418441d0c219f807cdca

Due to it's volume, this PR is not yet merged (thus you do not yet see the fix in master branch), but I'm positive that it will be soon.

PR #231, opened 2020-11-03
Add delete/lock/unlock REST api

OGAWAHirofumi:
This adds the following new REST api,

	/apis/rest/<item>/delete - for delete
	/apis/rest/<item>/lock   - for lock
	/apis/rest/<item>/unlock - for unlock

With this, client can do those easier without web form.

the following bepasty.txt (github doesn't support to attach .py or no ext) is a client what im now using. it supports text/x-bepasty-list as like directory and recursive upload/download etc., and using all REST apis include new apis in this PR.

im not sure, this client should go where - bepasty-server, bepasty-client-cli, or dont public

[bepasty.txt](https://github.com/bepasty/bepasty-server/files/5479329/bepasty.txt)

### DIFF
--- a/src/bepasty/apis/rest.py
+++ b/src/bepasty/apis/rest.py
@@ -15,7 +15,9 @@ from ..utils.name import ItemName
 from ..utils.permissions import CREATE, LIST, may
 from ..utils.upload import Upload, background_compute_hash
 from ..views.filelist import file_infos
+from ..views.delete import DeleteView
 from ..views.download import DownloadView
+from ..views.setkv import LockView, UnlockView
 
 
 # This wrappper handles exceptions in the REST api implementation.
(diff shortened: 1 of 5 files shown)


src/bepasty/tests/test_rest_server.py
```diff
@@ -783,6 +798,90 @@ def test_download_range(client_fixture):
...
+
+        # download locked item with admin (should success)
+        response = client.get(url.download, headers=add_auth('user', 'admin'))
+        check_data_response(response, metas[item_id], datas[item_id])
+
+        # delete locked item (should fail)
+        response = client.post(url.delete, headers=add_auth('user', 'full'))
+        check_err_response(response, 403)
+
+        # delete locked item with admin (should success)
```
ThomasWaldmann, line 876: should succeed

src/bepasty/tests/test_rest_server.py
```diff
@@ -806,11 +905,28 @@ def test_incomplete(client_fixture):
...
+
+    # unlock should error with incomplete
+    response = client.post(url.unlock, headers=add_auth('user', 'admin'))
+    check_err_response(response, 409)
+
+    # delete should error with incomplete
+    response = client.post(url.delete, headers=add_auth('user', 'full'))
+    check_err_response(response, 409)
+
+    # delete by admin with incomplete should success
```
ThomasWaldmann, line 930: should succeed

src/bepasty/apis/rest.py
```diff
@@ -239,6 +241,42 @@ def get(self, name):
         return super(ItemDetailView, self).get(name)
 
 
+class ItemDeleteView(DeleteView, RestBase):
+    def error(self, item, error):
+        raise Conflict(description=error)
+
+    def response(self, name):
+        return make_response('', {'Content-Type': 'application/json'})
```
ThomasWaldmann, line 249: how does this response look like exactly? is it valid json?

ThomasWaldmann: if this needs fixing, fix globally.

ThomasWaldmann (commented):
good PR :) some feedback, minor issues.

src/bepasty/tests/test_rest_server.py
OGAWAHirofumi: fixed

src/bepasty/tests/test_rest_server.py
OGAWAHirofumi: fixed

src/bepasty/apis/rest.py
OGAWAHirofumi: sure, response.json raise decode error. fixed by using '{}' instead of ''

ThomasWaldmann:
about your client: it looks like a completely new client, additionally to what we have in bepasty-client-cli.

maybe create a new repo, make a nice README.rst, point out what it can do and how it differs from the other client, use cases, etc.

if you like, we can have a new repo under bepasty organization on github.

OGAWAHirofumi:
my client can do all operations that provided by REST api. and already this is big different with bepasty-client-cli though, another big difference is, my client can list/upload/download directory (text/x-bepasty-list aware).

well, it is lazy to write README though, i may try simple one.

PR #233, opened 2020-11-03
Use json

OGAWAHirofumi:
this PR uses application/json for upload and error response.


ThomasWaldmann:
is this compatible with what bepasty-client-cli expects?

OGAWAHirofumi:
as far as i can tell, bepasty-client-cli doesn't check error or upload body (original is empty body). so i think not affect to it. however, of course, it is incompatible if someone uses upload/error response body if other rest api users.

ThomasWaldmann:
Thanks!

PR #236, opened 2020-11-08
Add optional python-magic support for application/octet-stream

OGAWAHirofumi:
This uses python-magic to improve mime type auto detection if user
didn't specify and filename couldn't detect.

To do it, make internal only metadata by "internal_meta[]". Then add
TYPE_HINT as internal metadata to remember to tell the pending state
of auto detection until file data is available.

[I'm not sure if there is better way to do this than internal_meta[] or not.]

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,6 +50,11 @@ setup(
         'xstatic-jquery-file-upload',
         'xstatic-pygments',
     ],
+    extras_require={
+        "magic": [
+            'python-magic'
+        ],
+    },
     classifiers=[
         'Development Status :: 4 - Beta',
         'Environment :: Web Environment',
--- a/src/bepasty/apis/rest.py
+++ b/src/bepasty/apis/rest.py
@@ -13,7 +13,7 @@ from ..utils.date_funcs import get_maxlife
 from ..utils.http import ContentRange, DownloadRange
 from ..utils.name import ItemName
 from ..utils.permissions import CREATE, LIST, may
-from ..utils.upload import Upload, background_compute_hash
+from ..utils.upload import Upload, filter_internal, background_compute_hash
 from ..views.filelist import file_infos
 from ..views.delete import DeleteView
 from ..views.download import DownloadView
@@ -204,7 +204,7 @@ class ItemUploadView(RestBase):
         for meta in file_infos():
             name = meta.pop(ID)
             ret[name] = {'uri': url_for('bepasty_apis.items_detail', name=name),
-                         'file-meta': meta}
+                         'file-meta': filter_internal(meta)}
         return jsonify(ret)
 
 
@@ -214,7 +214,7 @@ class ItemDetailView(DownloadView, RestBase):
 
     def response(self, item, name):
         return jsonify({'uri': url_for('bepasty_apis.items_detail', name=name),
-                        'file-meta': dict(item.meta)})
+                        'file-meta': filter_internal(item.meta)})
 
     @rest_errorhandler
     def get(self, name):
--- a/src/bepasty/config.py
+++ b/src/bepasty/config.py
@@ -70,6 +70,14 @@ class Config(object):
         '': 1 * 1000 * 1000,
     }
 
+    # Whether to use the python-magic module to guess a file's mime
+    # type by looking into its content (if the mime type can not be
+    # determined from the filename extension).
+    # NOTE:
+    # libmagic may have security issues, so maybe you should only use
+    # it if you trust all users with upload permissions ('create').
+    USE_PYTHON_MAGIC = False
+
     #: Define storage backend, choose from:
     #:
     #: - 'filesystem'
--- a/src/bepasty/constants.py
+++ b/src/bepasty/constants.py
@@ -1,5 +1,6 @@
 FILENAME = 'filename'
 TYPE = 'type'
+TYPE_HINT = 'type-hint'
 LOCKED = 'locked'
 SIZE = 'size'
 COMPLETE = 'complete'
@@ -12,3 +13,6 @@ FOREVER = -1
 
 # headers
 TRANSACTION_ID = 'Transaction-ID'  # keep in sync with bepasty-cli
+
+# used internally only
+internal_meta = [TYPE_HINT]
--- a/src/bepasty/tests/test_rest_server.py
+++ b/src/bepasty/tests/test_rest_server.py
@@ -897,6 +897,7 @@ def test_incomplete(client_fixture):
     response = _upload(client, data, token='full', filename=filename,
                        ftype=ftype, range_str=range_str)
     check_upload_response(response, 200)
+    assert len(response.headers[TRANSACTION_ID]) > 0
 
     # get incomplete item from list
     url = RestUrl()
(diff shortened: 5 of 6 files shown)


ThomasWaldmann:
Not sure I like this.

In general (and also specifically for libmagic), reading and parsing content is dangerous (esp. if done by complex C code).

Search for CVE and libmagic to see. There are some incidents from years ago and also one from 2019.

bepasty lets you freely select the mimetype manually in case autodetection from extension does not work.

ThomasWaldmann:
Also, for widespread cases of no detection or misdetection, we could also add or change mimetypes entries.

Noteable: there are 0 open tickets about no detection or misdetection.

OGAWAHirofumi:
> Not sure I like this.
> 
> In general (and also specifically for libmagic), reading and parsing content is dangerous (esp. if done by complex C code).
> 
> Search for CVE and libmagic to see. There are some incidents from years ago and also one from 2019.

then, just make USE_PYTHON_MAGIC=False by default. anyone care if only trusted user can upload?

> bepasty lets you freely select the mimetype manually in case autodetection from filename extension does not work.

impossible to specify with file uploads.

> Also, for widespread cases of no detection or misdetection, we could also add or change mimetypes entries.
> 
> Noteable: there are 0 open tickets about no detection or misdetection.

impossible to detect from filename. for example, Makefile

ThomasWaldmann:
Hmm, right, with some functionality, e.g. drag-and-drop upload, one can not give the mimetype.

So yeah, guess the default should be False and there should be a note right above that config line that parsing untrusted uploads via libmagic might be dangerous in case of security issues of libmagic.

src/bepasty/tests/test_rest_server.py
```diff
@@ -932,3 +933,60 @@ def test_incomplete(client_fixture):
...
+
+        # get detail to check python-magic auto detection
+        url = RestUrl(fid)
+        response = client.get(url.detail, headers=add_auth('user', 'full'))
+        check_json_response(response, None, check_data=False)
+
+        assert response.json['file-meta'] is not None
+        assert response.json['file-meta'][TYPE] != 'application/octet-stream'
+    except ImportError:
+        pass
```
ThomasWaldmann, line 992: This has a scoping issue: the block is rather wide and it might not be the ImportError you expected.

Better:

```
try:
    import magic
except ImportError:
    pass
else:
    # test code here
```

src/bepasty/utils/upload.py
```diff
@@ -5,6 +5,14 @@
 
 from flask import current_app
 
+try:
+    import magic
```
ThomasWaldmann, line 9: how about using `... as magic_module` here and `magic` instead of `mag`?

src/bepasty/tests/test_rest_server.py
ThomasWaldmann: for the case of importerror it could also do a `py.test.skip("skipping test, no python-magic installed")`.

src/bepasty/utils/upload.py
```diff
@@ -82,7 +92,10 @@ def meta_new(cls, item, input_size, input_filename, input_type,
             input_filename, storage_name, input_type, input_type_hint
         )
         item.meta[SIZE] = cls.filter_size(input_size)
-        item.meta[TYPE] = cls.filter_type(input_type, input_type_hint, input_filename)
+        ct, hint = cls.filter_type(input_type, input_type_hint, input_filename)
+        item.meta[TYPE] = ct
+        if hint and mag and current_app.config.get('USE_PYTHON_MAGIC'):
```
ThomasWaldmann, line 97: code works due to None being falsy, but slightly cleaner is: `... .get('USE_PYTHON_MAGIC', False):`

src/bepasty/utils/upload.py
```diff
@@ -72,8 +82,8 @@ def filter_type(cls, ct, ct_hint, filename=None):
         if not ct and filename:
             ct, encoding = mimetypes.guess_type(filename)
         if not ct:
-            return ct_hint
-        return cls._type_re.sub('', ct)[:50]
+            return ct_hint, True
+        return cls._type_re.sub('', ct)[:50], False
```
ThomasWaldmann, line 86: please document the meaning of the return value elements.

src/bepasty/utils/upload.py
```diff
@@ -92,6 +105,12 @@ def meta_new(cls, item, input_size, input_filename, input_type,
 
     @classmethod
     def meta_complete(cls, item, file_hash):
+        # update TYPE by python-magic if not decided yet
+        if item.meta.get(TYPE_HINT):
+            del item.meta[TYPE_HINT]
```
ThomasWaldmann, line 110: similar here. you could also use `.pop(KEY, False)`

src/bepasty/apis/rest.py
```diff
@@ -214,7 +214,7 @@ def err_incomplete(self, item, error):
 
     def response(self, item, name):
         return jsonify({'uri': url_for('bepasty_apis.items_detail', name=name),
-                        'file-meta': dict(item.meta)})
+                        'file-meta': filter_internal(dict(item.meta))})
```
ThomasWaldmann, line 217: guess you don't need the `dict()` here, `filter_meta` creates a new dict.

src/bepasty/utils/upload.py
ThomasWaldmann: guess you could also **not** do this conditional here and just write hint to the metadata, no matter whether True or False.

then, when the item is completed, you remove the hint again anyway and you can do the conditional there only.

src/bepasty/config.py
```diff
@@ -70,6 +70,12 @@ class Config(object):
         '': 1 * 1000 * 1000,
     }
 
+    #: Use python-magic module to guess mime type additionally, if
+    #: mime type is not determined from filename.
+    #: NOTE: libmagic may have security hole, so it may not be good to
+    #: enable this for publicly uploadable server
+    USE_PYTHON_MAGIC = False
```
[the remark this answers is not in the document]
OGAWAHirofumi: changed

setup.py
```diff
@@ -40,6 +40,7 @@
     install_requires=[
         'Flask',
         'Pygments',
+        # 'python-magic',
```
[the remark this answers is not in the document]
OGAWAHirofumi: hmm, i dont know what is it. well, i will try to do

src/bepasty/constants.py
```diff
@@ -12,3 +13,6 @@
 
 # headers
 TRANSACTION_ID = 'Transaction-ID'  # keep in sync with bepasty-cli
+
+# used internal only
```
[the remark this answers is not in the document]
OGAWAHirofumi: fixed

src/bepasty/tests/test_rest_server.py
```diff
@@ -932,3 +933,60 @@ def test_incomplete(client_fixture):
...
+    # 'type-hint' should be invisible
+    assert meta['file-meta'] is not None
+    assert meta['file-meta'].get('type-hint') is None
+
+    # 'type' is 'application/octet-stream' for now
+    assert meta['file-meta'][TYPE] == 'application/octet-stream'
+
+    try:
+        import magic
+        assert magic is not None
```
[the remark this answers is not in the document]
OGAWAHirofumi: this is suppress unused warning from flake8

src/bepasty/tests/test_rest_server.py
```diff
@@ -932,3 +933,60 @@ def test_incomplete(client_fixture):
...
+                           trans_id=trans_id)
+        check_upload_response(response)
+
+        # get detail to check python-magic auto detection
+        url = RestUrl(fid)
+        response = client.get(url.detail, headers=add_auth('user', 'full'))
+        check_json_response(response, None, check_data=False)
+
+        assert response.json['file-meta'] is not None
+        assert response.json['file-meta'][TYPE] != 'application/octet-stream'
```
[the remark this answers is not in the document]
OGAWAHirofumi: in theory, it can be depending on libmagic version...

src/bepasty/utils/upload.py
OGAWAHirofumi: fixed

src/bepasty/utils/upload.py
OGAWAHirofumi: hm, im not sure i got comment correctly though. if hint is False, we dont need to add TYPE_HINT field...

src/bepasty/tests/test_rest_server.py
OGAWAHirofumi: fixed

src/bepasty/utils/upload.py
OGAWAHirofumi: added following comment

```python
        # Return value:
        # tuple[0] - content-type string
        # tuple[1] - whether tuple[0] is hint or not
        #            True:  content-type is just a hint
        #            False: content-type was specified by user
```

src/bepasty/utils/upload.py
OGAWAHirofumi: fixed

src/bepasty/apis/rest.py
OGAWAHirofumi: looks like work without dict()

setup.py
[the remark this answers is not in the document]
ThomasWaldmann: https://creatronix.de/pip-optional-dependencies/

src/bepasty/tests/test_rest_server.py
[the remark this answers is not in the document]
ThomasWaldmann: ah!

src/bepasty/tests/test_rest_server.py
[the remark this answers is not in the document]
ThomasWaldmann: right, if that happens, we can fix it.

src/bepasty/utils/upload.py
ThomasWaldmann: yes. but as you remove it later anyway, it would not do harm and simplify the code.

src/bepasty/utils/upload.py
```diff
@@ -71,9 +81,14 @@ def filter_type(cls, ct, ct_hint, filename=None):
         """
         if not ct and filename:
             ct, encoding = mimetypes.guess_type(filename)
+        # Return value:
+        # tuple[0] - content-type string
+        # tuple[1] - whether tuple[0] is hint or not
+        #            True:  content-type is just a hint
+        #            False: content-type is not a hint, was specified by user
```
ThomasWaldmann, line 88: move to docstring of method?

setup.py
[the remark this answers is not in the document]
OGAWAHirofumi: added following

```python
    extras_require={
        "magic": [
            'python-magic'
        ],
    },
```

src/bepasty/tests/test_rest_server.py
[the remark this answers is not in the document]
OGAWAHirofumi: ok, fixed

src/bepasty/utils/upload.py
OGAWAHirofumi: right, however this is on-disk data. so it would be better to keep as possible as smaller?

OGAWAHirofumi:
changed

- move filter_type comment to doc string
- test 'text/x-python'

setup.py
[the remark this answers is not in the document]
ThomasWaldmann: thanks

src/bepasty/utils/upload.py
ThomasWaldmann: it's only temporarily larger and only a few bytes, so it doesn't matter.

ThomasWaldmann: so do it however you prefer, i am ok with both ways. tell me when finished so i can merge this.

OGAWAHirofumi: ok, i changed to store unconditionally

OGAWAHirofumi:
changed

- store/remove TYPE_HINT unconditionally

prev,

- move filter_type comment to doc string
- test 'text/x-python'

ThomasWaldmann:
thanks!

PR #238, opened 2020-11-09
Disallow the old cookie for PERMISSIONS in old config

OGAWAHirofumi:
This changes the content of session cookie from PERMISSIONS string to
password.

Because, without this change, even if PERMISSIONS in server config was
changed, PERMISSIONS in cookie are still used until the cookie is
removed from client side.

Also possibly more secure even if SECRET_KEY was leaked? Because user
can't add PERMISSIONS by modifying the cookie.

### DIFF
--- a/src/bepasty/app.py
+++ b/src/bepasty/app.py
@@ -1,5 +1,6 @@
 import os
 import time
+import hashlib
 
 from flask import (
     Flask,
@@ -45,12 +46,33 @@ class PrefixMiddleware(object):
             return ['This URL does not belong to the bepasty app.'.encode()]
 
 
+def setup_secret_key(app):
+    """
+    The secret key is used to sign cookies and cookies not signed with the
+    current secret key are considered invalid.
+
+    Here, we amend the configured secret key, so it depends on some
+    other config values. Changing any of these values will change the
+    computed secret key (and thus invalidate all previously made
+    cookies).
+
+    Currently supported secret-changing config values: PERMISSIONS
+    """
+    # if app.config['SECRET_KEY'] is empty, keep as NullSession
+    if app.config['SECRET_KEY']:
+        perms = sorted(k + v for k, v in app.config['PERMISSIONS'].items())
+        perms = ''.join(perms).encode()
+        app.config['SECRET_KEY'] += hashlib.sha256(perms).hexdigest()
+
+
 def create_app():
     app = Flask(__name__)
 
     app.config.from_object('bepasty.config.Config')
     if os.environ.get('BEPASTY_CONFIG'):
         app.config.from_envvar('BEPASTY_CONFIG')
+
+    setup_secret_key(app)
 
     prefix = app.config.get('APP_BASE_PATH')
     if prefix is not None:
--- a/src/bepasty/tests/test_app.py
+++ b/src/bepasty/tests/test_app.py
@@ -1,0 +1,26 @@
+#
+# app tests
+#
+
+from ..app import create_app
+from ..config import Config
+
+
+def test_secret_key():
+    Config.PERMISSIONS = {
+        'admin': 'admin,list,create,read,delete',
+        'full': 'list,create,read,delete',
+        'none': '',
+    }
+    Config.SECRET_KEY = 'secret'
+
+    app = create_app()
+    secret_key = app.config['SECRET_KEY']
+    assert len(secret_key) > len(Config.SECRET_KEY)
+
+    Config.PERMISSIONS = {
+        'admin': 'admin,list,create,read,delete',
+        'none': '',
+    }
+    app = create_app()
+    assert app.config['SECRET_KEY'] != secret_key


ThomasWaldmann:
Not sure what to think of that.

You're right about outdated permissions being alive while the cookie is valid, that's bad.

The client is unable to make up its faked own cookie though as the secure cookie is signed server-side using a key (and we can be rather sure the client does not have the key - at least if the server admin is not doing anything stupid).

OTOH, the cookie is not encrypted (IIRC), so it is not great sending the bepasty "login secret" over the wire with each http request.

Does not matter with https though as we have an encrypted tunnel, but with http (otoh, if http is used it is just about the secret being transmitted rarely vs. the secret being transmitted all the time, which is different, but not fundamentally different as you could sniff it in both cases).

Also, the cookie (and thus the bepasty "login secret") is stored on the client machine. The cookie has some lifetime (IIRC), but the login secret might be valid for a long time.

OGAWAHirofumi:
ok, i checked how generate cookie string. it looks like base64(compress(sha512(salt))).
looks like no encrypt if i read correctly.

so, another idea is store PERMISSIONS config id (cookie) to http cookie. config id is generated from sorted PERMISSIONS data. this may be detects obsoleted PERMISSIONS in http cookie.

ThomasWaldmann:
hmm, maybe it is simpler to just set (update) the cookie with fresh permissions from config with every request?

OGAWAHirofumi:
> hmm, maybe it is simpler to just set (update) the cookie with fresh permissions from config with every request?

there is no password in the cookie. so we don't know which entry in PERMISSIONS.

OGAWAHirofumi:
New attempt

Without this change, even if PERMISSIONS in server config was changed
(for example, dropped password), PERMISSIONS in cookie are still used.
until the cookie is removed from client side. So client can login with
old PERMISSIONS.

src/bepasty/utils/permissions.py
```diff
@@ -21,6 +24,40 @@
...
+    old config, this makes unique id for current config.
+
+    NOTE: for now, only permissions is checked
+    """
+    global session_generation_cache
+
+    if session_generation_cache is None:
+        perms = []
+        for k, v in current_app.config['PERMISSIONS'].items():
+            perms.append(k + v)
```
ThomasWaldmann, line 50: perms = sorted(k + v for k, v in current_app.config['PERMISSIONS'].items())

src/bepasty/utils/permissions.py
```diff
@@ -21,6 +24,40 @@
...
+def session_generation():
+    """
+    to detect the configuration change, and make obsolete session for
+    old config, this makes unique id for current config.
+
+    NOTE: for now, only permissions is checked
+    """
+    global session_generation_cache
+
+    if session_generation_cache is None:
```
ThomasWaldmann, line 47: s/session_generation_cache/permissions_fingerprint/ ?

src/bepasty/utils/permissions.py
```diff
@@ -11,6 +13,7 @@
 
 # key in the session:
 PERMISSIONS = 'permissions'
+GENERATION = 'generation'
```
ThomasWaldmann, line 16: FINGERPRINT?

ThomasWaldmann (commented):
yeah, guess this would work.

i just had another idea:

if we just document (e.g. right about the PERMISSIONS config entry) that when changing the entries one also should change the server side secret, this would automatically invalidate all sessions and solve this issue without any addtl. code.

OGAWAHirofumi:
> i just had another idea:
> 
> if we just document (e.g. right about the PERMISSIONS config entry) that when changing the entries one also should change the server side secret, this would automatically invalidate all sessions and solve this issue without any addtl. code.

sure, further idea. this patch do it automatically by adding hash to SECRET_KEY.

src/bepasty/app.py
```diff
@@ -45,13 +46,29 @@ def __call__(self, environ, start_response):
...
+    to detect the configuration change, and make obsolete session for
+    old config, this add unique id to SECRET_KEY additionally.
+
+    NOTE: for now, only permissions is checked
+    """
+    # if app.config['SECRET_KEY'] is empty, keep as NullSession
+    if app.config['SECRET_KEY']:
+        perms = sorted(k + v for k, v in app.config['PERMISSIONS'].items())
+        perms = ''.join(perms).encode()
+        app.config['SECRET_KEY'] += hashlib.sha256(perms).hexdigest()
```
ThomasWaldmann, line 60: ah, great idea :-)

src/bepasty/app.py
```diff
@@ -45,13 +46,29 @@ def __call__(self, environ, start_response):
...
+
+
 def create_app():
     app = Flask(__name__)
 
     app.config.from_object('bepasty.config.Config')
     if os.environ.get('BEPASTY_CONFIG'):
         app.config.from_envvar('BEPASTY_CONFIG')
 
+    setup_secret_key(app)
```
ThomasWaldmann, line 70: hopefully this is early enough, before flask actually does anything with the secret.

src/bepasty/tests/test_rest_server.py
```diff
@@ -139,6 +139,26 @@ def unlock(self):
...
+    assert secret_key > Config.SECRET_KEY
+
+    Config.PERMISSIONS = {
+        'admin': 'admin,list,create,read,delete',
+        'none': '',
+    }
+    app = create_app()
+    assert app.config['SECRET_KEY'] != secret_key
+
+
```
ThomasWaldmann, line 161: how is that related to `test_rest_server`? does it maybe fit better into another testing module?

src/bepasty/tests/test_rest_server.py
```diff
@@ -139,6 +139,26 @@ def unlock(self):
...
+    Config.PERMISSIONS = {
+        'admin': 'admin,list,create,read,delete',
+        'full': 'list,create,read,delete',
+        'none': '',
+    }
+    Config.SECRET_KEY = 'secret'
+
+    app = create_app()
+    secret_key = app.config['SECRET_KEY']
+    assert secret_key > Config.SECRET_KEY
```
ThomasWaldmann, line 152: did you really mean sort order or did you forget `len(...) > len(...)`?

src/bepasty/app.py
OGAWAHirofumi: thanks :)

src/bepasty/app.py
OGAWAHirofumi: ok

src/bepasty/tests/test_rest_server.py
OGAWAHirofumi: ok

src/bepasty/tests/test_rest_server.py
OGAWAHirofumi: ouch, fixed

OGAWAHirofumi:
changes

- use test_app.py for test
- fix typo of len()

src/bepasty/app.py
```diff
@@ -45,13 +46,29 @@ def __call__(self, environ, start_response):
             return ['This URL does not belong to the bepasty app.'.encode()]
 
 
+def setup_secret_key(app):
+    """
+    to detect the configuration change, and make obsolete session for
+    old config, this add unique id to SECRET_KEY additionally.
+
+    NOTE: for now, only permissions is checked
```
ThomasWaldmann, line 54: Maybe rephrase this:

The secret key is used to sign cookies and cookies not signed with the current secret key are considered invalid.

Here, we amend the configured secret key, so it depends on some other config values. Changing any of these values will change the computed secret key (and thus invalidate all previously made cookies).

Currently supported secret-changing config values: PERMISSIONS

ThomasWaldmann (commented):
suggestion for improved docstring

other than that, guess we are ready to merge!

OGAWAHirofumi:
changes

- adjust comment for setup_secret_key()
prev

- use test_app.py for test
- fix typo of len()

ThomasWaldmann:
\o/

PR #240, opened 2020-11-10
Add edit dialog to modify metadata (for now)

OGAWAHirofumi:
This adds feature to modify filename and mime-type (at least for now) of uploaded item.
And both of web UI and REST api are supported, and add new EDIT permission to allow
this operations.

the following screenshots are including edit dialog,

[screenshots.tar.gz](https://github.com/bepasty/bepasty-server/files/5516929/screenshots.tar.gz)

[this would be last one what i wanted for now]

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ setup(
         'Pygments',
         'xstatic',
         'XStatic-asciinema-player',
-        'xstatic-bootbox',
+        'xstatic-bootbox>=5.4.0',
         'xstatic-bootstrap>=4.0.0.0,<5.0.0.0',
         'xstatic-font-awesome',
         'xstatic-jquery',
--- a/src/bepasty/static/app/js/utils.js
+++ b/src/bepasty/static/app/js/utils.js
@@ -8,6 +8,37 @@ $(function () {
         });
     });
 
+    // Show a modify dialog box when trying to edit metadata.
+    $("#modify-btn").click(function() {
+        var form_name = "modify-frm"
+        var hidden_name_id = "#hidden-" + form_name
+        // Read form template from html
+        var modal_form = $(hidden_name_id).html();
+        var modal_title = $(hidden_name_id).attr('modalTitle');
+        var modal_focus = $(hidden_name_id).attr('modalFocus');
+        // A bit of a hack to avoid implementing this from scratch
+        // using .dialog().
+        var box = bootbox.confirm({
+            title: modal_title,
+            message: modal_form,
+            centerVertical: true,
+            onShown: function(e) {
+                if (modal_focus) {
+                    $(this).find("#" + modal_focus).trigger('focus');
+                }
+                // Support jquery-ui autocomplete
+                contenttype_autocomplete(this)
+            },
+            callback: function(result) {
+                // Please note that this is not called when hitting
+                // the Enter key on the input box.
+                if (result == true) {
+                    $(this).find("#" + form_name).submit();
+                }
+            }
+        });
+    });
+
     // Bind on click event to all line number anchor tags
     $('td.linenos a').on('click', function(e) {
         remove_highlights();
--- a/src/bepasty/templates/display.html
+++ b/src/bepasty/templates/display.html
@@ -20,7 +20,7 @@
                     <span class="fa fa-asterisk"></span> Inline
                 </a>
             {% endif %}
-            {% if may(CREATE) %}
+            {% if may(MODIFY) %}
                 <div id="hidden-modify-frm" class="d-none" modalTitle="Modify Metadata" modalFocus="filename">
                     <!-- modify form that used by utils.js -->
                     <form id="modify-frm" action="{{ url_for('bepasty.modify', name=name) }}" method="post">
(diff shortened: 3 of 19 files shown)


src/bepasty/static/app/js/utils.js
```diff
@@ -8,6 +8,35 @@ $(function () {
...
+            centerVertical: true,
+            onShown: function(e) {
+                if (modal_focus) {
+                    $(this).find("#" + modal_focus).trigger('focus');
+                }
+                // Support jquery-ui autocomplete
+                contenttype_autocomplete(this)
+            },
+            callback: function(result) {
+                // Note, if pushed Enter key on input boxes, this is not called
```
ThomasWaldmann, line 32: please note that this is not called when hitting the Enter key on the input box.

src/bepasty/static/app/js/utils.js
```diff
@@ -8,6 +8,35 @@ $(function () {
         });
     });
 
+    // Show a edit dialog box when trying to edit metadata
```
ThomasWaldmann, line 11: Show an edit dialog box when trying to edit metadata.

src/bepasty/templates/display.html
```diff
@@ -76,4 +101,12 @@ <h1>
...
 <script src="{{ url_for('bepasty.xstatic', name='asciinema_player', filename='asciinema-player.js') }}" type="text/javascript"></script>
 <script src="{{ url_for('static', filename='app/js/utils.js') }}" type="text/javascript"></script>
+{% if may(EDIT) %}
+<script>
+    <!-- function that used by utils.js -->
+    function contenttype_autocomplete(modal_box) {
+        {{ utils.contenttype_autocomplete('$(modal_box).find("#contenttype")', contenttypes) }}
+    };
+</script>
+{% endif %}
```
ThomasWaldmann, line 111: could this function definition just be in utils.js and not inlined here?

guess its definition does not need to depend on "may edit"?

ThomasWaldmann (commented):
nice. some small issues i found.

also: i was thinking about whether this should be called "modify" rather than "edit".

when thinking of "edit", a lot of people rather think of editing a text (== the item content) rather than the metadata.

but considering we maybe never should implement changing the content anyway (just think of 2 different users with edit permissions changing the other's stuff), guess it does not really matter.

changing the mimetime and the filename has way less potential of abuse and also is something that users might need to fix after upload, we also have a ticket about that.

ThomasWaldmann:
btw, as a general note:

i'ld like to do a release soon. considering you've added a lot of nice functionality, we should get that out to the bepasty users.

this needs a bit of planning / prioritizing or delaying stuff, see the milestones.
if you think something in the milestones should be adjusted, just tell in the respective ticket.
if something is missing, open a ticket.

setup.py
```diff
@@ -42,7 +42,7 @@
         'Pygments',
         'xstatic',
         'XStatic-asciinema-player',
-        'xstatic-bootbox',
+        'xstatic-bootbox>=5.4.0',
```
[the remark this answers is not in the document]
OGAWAHirofumi: bootbox document says onShown property was added at 5.4.0.
so it is 5.4.0.

src/bepasty/static/app/js/utils.js
```diff
@@ -8,6 +8,35 @@ $(function () {
...
+    $("#edit-btn").click(function() {
+        var form_name = "edit-frm"
+        var hidden_name_id = "#hidden-" + form_name
+        // Read form template from html
+        var modal_form = $(hidden_name_id).html();
+        var modal_title = $(hidden_name_id).attr('modalTitle');
+        var modal_focus = $(hidden_name_id).attr('modalFocus');
+        // A bit hack, to avoid implement with .dailog() from scratch
```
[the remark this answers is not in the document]
OGAWAHirofumi: fixed

src/bepasty/static/app/js/utils.js
OGAWAHirofumi: fixed

src/bepasty/static/app/js/utils.js
OGAWAHirofumi: fixed. with "an edit" => "a modify"

src/bepasty/templates/display.html
OGAWAHirofumi: this extends contenttypes list from python. so maybe more complex to pass it to javascript

OGAWAHirofumi:
> btw, as a general note:
> 
> i'ld like to do a release soon. considering you've added a lot of nice functionality, we should get that out to the bepasty users.
> 
> this needs a bit of planning / prioritizing or delaying stuff, see the milestones.
> if you think something in the milestones should be adjusted, just tell in the respective ticket.
> if something is missing, open a ticket.

hm, IMO, i would like to add this one to next release if possible. however, it is you decide...

setup.py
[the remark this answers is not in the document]
ThomasWaldmann: considering that 5.4.0 is out since almost 1 year, guess we do not need that for now.

so, keep this as is.

src/bepasty/templates/display.html
ThomasWaldmann: a right, it's calling py code via templating... - keep it as is.

ThomasWaldmann:
note: i've add this PR to 0.6.0 milestone (== next release, soon).

ThomasWaldmann:
Tell me when you're finished here, then I'll merge this.

ThomasWaldmann:
BTW, did you also add an icon for "modify" (for the "Permissions: ..." display at top of the theme)?

OGAWAHirofumi:
changes

- comments fixes
- edit => modify

updated screenshots are here

[screenshots.tar.gz](https://github.com/bepasty/bepasty-server/files/5519251/screenshots.tar.gz)

OGAWAHirofumi:
> BTW, did you also add an icon for "modify" (for the "Permissions: ..." display at top of the theme)?

yes, fa-edit, above screenshot should show how look

ThomasWaldmann:
Is this finished?

OGAWAHirofumi:
> Is this finished?

ah, yes.

ThomasWaldmann:
Thanks!

PR #256, opened 2020-12-11
don't crash on ``('JSONBareObject', (), (), ())``

mfa:


mfa:
testfail on py3.9 is strange. my server runs on py3.9 and everything worked after this patch.

ThomasWaldmann:
@mfa thanks for the PR!

Let's see if the fail is reproducible, I just restarted the testrun.

ThomasWaldmann:
OK, works now.

So, that JSONBareObject is a recent addition to pygments?

PR #259, opened 2020-12-20
Update install-tutorial.rst

oijn:
Updated tutorial instructions for latest Debian+Python3.

### DIFF
--- a/docs/source/install-tutorial.rst
+++ b/docs/source/install-tutorial.rst
@@ -3,19 +3,32 @@
 Installation tutorial with Debian, NGinx and gunicorn
 =====================================================
 
-preliminary packages:
+Following this tutorial you'll have a running Debian+Nginx+gunicorn+bepasty (HTTP) instance using Python3 on a minimally installed Debian server.  It is strongly recommended to use HTTPS for Internet facing services - configuring Nginx for HTTPS is left to the reader.
+
+After installing a minimal Debian 10 server.
+
+Install preliminary required packages:
 
 ::
 
-  apt-get install build-essential nginx supervisor python-dev git-core python-pip python-virtualenv
+  apt-get install nginx git-core python3-dev python3-pip virtualenv python3-virtualenv
 
 
-commands to run
+Some commands need to be run as root - others as the user bepasty.
+
+Commands to run as root:
 
 ::
 
   # add user bepasty to system
-  adduser bepasty
+  adduser bepasty --disabled-login
+  # add path to bepasty/.bash_rc
+  echo 'export PATH="/home/bepasty/.local/bin:$PATH"' >> /home/bepasty/.bash_rc
+  
+Commands to run as user bepasty:
+
+::
+
   # change to user bepasty
   sudo su - bepasty
   # clone repository from github
@@ -30,24 +43,36 @@ commands to run
   . bin/activate
   cd repo
   # install bepasty and requirements
-  pip install -e .
+  pip3 install -e .
   # add gunicorn and gevent for hosting
-  pip install gunicorn gevent
+  pip3 install gunicorn gevent
 
-config file for bepasty -- ``/home/bepasty/bepasty.conf``:
+Config file for bepasty -- ``/home/bepasty/bepasty.conf``:
 
 Copy ``src/bepasty/config.py`` to ``/home/bepasty/bepasty.conf`` first,
 remove the ``class Config`` and remove all indents in the file.
 The comments can be removed too, if you feel the need to.
-At last modify these two configs variables:
+
+An easy way to do this in one go:
 
 ::
 
+  # create trim alias
+  alias trim="awk '{\$1=\$1};1'"
+  # remove the top 11 lines + trim spaces at beginning of lines
+  trim src/bepasty/config.py | tail -n +11 > /home/bepasty/bepasty.conf
+
+
+At least modify these four config variables.  
+
+::
+
+  SITENAME = 'bepasty.example.org'
   STORAGE = 'filesystem'
   STORAGE_FILESYSTEM_DIRECTORY = '/home/bepasty/storage/'
+  SESSION_COOKIE_SECURE = False
 
-
-add this content to ``/home/bepasty/bin/gunicorn_bepasty``:
+Add this content to ``/home/bepasty/bin/gunicorn_bepasty``:
 
 ::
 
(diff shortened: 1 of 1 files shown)


docs/source/install-tutorial.rst
```diff
@@ -30,24 +43,36 @@ commands to run
...
-At last modify these two configs variables:
+
+An easy way to do this in one go:
+
+::
+
+  # create trim alias
+  alias trim="awk '{\$1=\$1};1'"
+  # remove the top 11 lines + trim spaces at beginning of lines
+  trim src/bepasty/config.py | tail -n +11 > /home/bepasty/bepasty.conf
```
ThomasWaldmann, line 63: please do not invent such tools for an install tutorial. just describe what needs to be done.

docs/source/install-tutorial.rst
```diff
@@ -58,6 +83,7 @@ add this content to ``/home/bepasty/bin/gunicorn_bepasty``:
   SOCKFILE=$HOME/gunicorn.sock  # we will communicate using this unix socket
   PIDFILE=$HOME/gunicorn.pid
   NUM_WORKERS=3                 # how many worker processes should Gunicorn spawn
+  export PATH="/home/bepasty/.local/bin:$PATH"  #include gunicorn path for exec
```
ThomasWaldmann, line 86: always have 2 spaces (at least) before and 1 space after `#`.

docs/source/install-tutorial.rst
```diff
@@ -85,7 +115,7 @@ A nginx configuration i.e. in ``/etc/nginx/conf.d/bepasty.conf``:
   server {
     listen 80;
     #listen [::]:80; #uncomment this if your server supports IPv6
-    server_name paste.example.org;  # <-- add your domainname here
+    server_name bepasty.example.org;  # <-- add your domainname here
```
ThomasWaldmann, line 118: thanks for wanting to mention bepasty, but paste or pastebin is the more well-known name for that kind of service.

docs/source/install-tutorial.rst
```diff
@@ -115,3 +145,36 @@ Supervisord config i.e. in ``/etc/supervisor/conf.d/bepasty.conf``:
...
+
+Places to look when troubleshooting:
+
+::
+
+  journalctl -u nginx.service
+  journalctl -u supervisor.service
+  tail /home/bepasty/logs/nginx-access.log
+  tail /home/bepasty/logs/nginx-error.log
+  tail /home/bepasty/logs/gunicorn_supervisor.log
```
ThomasWaldmann, line 161: never use tail to output logs to your console, that is dangerous. better: `less +F filename`

docs/source/install-tutorial.rst
```diff
@@ -115,3 +145,36 @@ Supervisord config i.e. in ``/etc/supervisor/conf.d/bepasty.conf``:
...
+  ``'user_secret': 'correct,list,of,permissions',``
+  any missing ``'' : '',`` will result in the bepasty server not starting.
+  
+Tips & Tricks:
+
+* The user secret in ``bepasty/config.py`` needs to be secure - an easy way 
+  to create secure random character strings on the command line is:
+  ``tr -dc A-Za-z0-9 </dev/urandom | head -c 10 ; echo ''``
+  this can also be used to create a long random secure SECRET_KEY value:
+  ``tr -dc 'A-Za-z0-9!"#$%&'\''()*+,-./:;<=>?@[\]^_`{|}~' </dev/urandom | head -c 255  ; echo ''``
```
ThomasWaldmann, line 180: Maybe use same command for both and just tell to use a longer secret for the SECRET_KEY.

docs/source/install-tutorial.rst
```diff
@@ -3,19 +3,32 @@
...
+Following this tutorial you'll have a running Debian+Nginx+gunicorn+bepasty (HTTP) instance using Python3 on a minimally installed Debian server.  It is strongly recommended to use HTTPS for Internet facing services - configuring Nginx for HTTPS is left to the reader.
+
+After installing a minimal Debian 10 server.
+
+Install preliminary required packages:
 
 ::
 
-  apt-get install build-essential nginx supervisor python-dev git-core python-pip python-virtualenv
+  apt-get install nginx git-core python3-dev python3-pip virtualenv python3-virtualenv
```
ThomasWaldmann, line 14: so, you did not need build-essential? is it installed on your base system already?

what's up with the supervisor stuff?

ThomasWaldmann:
@oijn did you see my feedback?

ThomasWaldmann:
BTW, I'll make a 1.0 release soon, maybe we can get this PR merged before that.

docs/source/install-tutorial.rst
oijn: Maybe we can agree to put the recipe in the Tips & Tricks instead?

docs/source/install-tutorial.rst
oijn: OK - understood.

docs/source/install-tutorial.rst
oijn: I think this needs to be consistent throughout the tutorial.  In the tutorial section that modifies /home/bepasty/bepasty.conf
SITENAME = 'bepasty.example.org'

Therefore, in my humble opinion - whatever is used pasty.example.org or bepasty.example.org has to be consistent so that those that are new to bepasty are not confused.

docs/source/install-tutorial.rst
oijn: OK - point taken.  Thanks for pointing this out.

docs/source/install-tutorial.rst
oijn: The difference is that the first example only uses characters that can be typed on any language keyboard.  The second uses almost the full list of the suggested characters for strong keys - of which may be difficult to enter on different keyboards.
I'm not against the idea and can modify as per your suggestion.

docs/source/install-tutorial.rst
oijn: Good spot regarding the missing supervisor.

On Debian 10 - no need for build-essential as anything that is required by the python/pip/etc.. packages are pulled in anyway.

I have now installed twice more to verify and I have confirmed that all you need is:
apt-get install git-core python3-dev python3-pip virtualenv python3-virtualenv nginx supervisor

oijn:
> @oijn did you see my feedback?

Sorry - I was meaning to get back to this - have now just done so.

I have also re-run the tutorial two extra times and have changed the order of some points and modified some.

ThomasWaldmann:
superseded by #264

PR #262, opened 2021-01-08
Docker Image

foosinn:
This provides a Docker image for bepasty.

Goals:

* run as unprivileged user
* no write access to config
* tries to be simple
* basic configuration via environment variables
* extended configuration via file mount

I'm not entirely sure if a Docker image is wanted. Neither found one in the repository nor the issues.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,30 @@
+FROM python:3.11-slim AS builder
+
+RUN true \
+  && mkdir /app \
+  && apt update \
+  && apt install --no-install-recommends -y git \
+  && python -m venv /app/env \
+  && /app/env/bin/pip install wheel gunicorn
+
+ADD . /usr/local/src/bepasty
+RUN /app/env/bin/pip install /usr/local/src/bepasty
+
+# ---
+
+FROM python:3.11-slim
+
+RUN true \
+  && mkdir /app \
+  && adduser -u 17357 --system --home /app/data --disabled-login --disabled-password paste
+
+ADD docker/init.sh /usr/local/bin/init.sh
+COPY --from=builder /app/env /app/env
+
+ENV WORKERS 4
+ENV LISTEN 0.0.0.0:8000
+ENV BEPASTY_CONFIG /etc/bepasty.conf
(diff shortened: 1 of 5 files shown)


ThomasWaldmann:
can somebody who is familiar with docker please review this?

duchenpaul:
This is nice, please merge this PR

ThomasWaldmann:
@duchenpaul so you practically tried it?

docker/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Quickstart
+
+```sh
+# build from the repository root
+docker build -f docker/Dockerfile -t bepasty .
+
+# create datadir
+mkdir data
+sudo chown 17357:0 data
```
ThomasWaldmann, line 20: the chown uid is not the same as the uid the comment (line 9) talks about.

also unclear to me: where does that uid come from?

docker/Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM python:3.9-slim AS builder
```
ThomasWaldmann, line 1: could something with 3.11 be used now?

docker/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+  && apt install --no-install-recommends -y git \
+  && python -m venv /app/env \
+  && /app/env/bin/pip install wheel gunicorn
+
+ADD . /usr/local/src/bepasty
+RUN /app/env/bin/pip install /usr/local/src/bepasty
+
+# ---
+
+FROM python:3.9-slim
```
ThomasWaldmann, line 15: 3.11?

ThomasWaldmann (commented):
had a quick look, some feedback/questions.

docker/README.md
foosinn: thanks! fixed

docker/Dockerfile
foosinn: updated but for now untested. should be able to have a look at the end of the week

Ilosariph:
A docker image would bi amazing, is there any plan for this?

PR #264, opened 2021-01-10
Update install-tutorial.rst

oijn:
I have updated the tutorial and verified by installing from a new Debian 10 minimal server install twice.

I have also (tried) to create a new pull request and answered the comments.

### DIFF
--- a/docs/source/install-tutorial.rst
+++ b/docs/source/install-tutorial.rst
@@ -3,21 +3,36 @@
 Installation tutorial with Debian, NGinx and gunicorn
 =====================================================
 
-preliminary packages:
+Following this tutorial you'll have a running Debian+Nginx+gunicorn+bepasty (HTTP) instance using Python3 on a minimally installed Debian server.  It is strongly recommended to use HTTPS for Internet facing services - configuring Nginx for HTTPS is left to the reader.
+
+After installing a minimal Debian 10 server.
+
+Install preliminary required packages:
 
 ::
 
-  apt-get install build-essential nginx supervisor python-dev git-core python-pip python-virtualenv
+  apt-get install git-core python3-dev python3-pip virtualenv python3-virtualenv nginx supervisor 
 
 
-commands to run
+Some commands need to be run as root - others as the user bepasty.
+
+Commands to run as root:
 
 ::
 
   # add user bepasty to system
-  adduser bepasty
+  adduser bepasty --disabled-login
+  
+Commands to run as user bepasty:
+
+::
+
   # change to user bepasty
-  sudo su - bepasty
+  su - bepasty
+  # add path to bepasty/.bashrc
+  echo "export PATH=\"/home/bepasty/.local/bin:\$PATH\"" >> /home/bepasty/.bashrc
+  # reload .bashrc
+  source ~/.bashrc
   # clone repository from github
   git clone https://github.com/bepasty/bepasty-server.git repo
   # create folder for storage
@@ -30,24 +45,27 @@ commands to run
   . bin/activate
   cd repo
   # install bepasty and requirements
-  pip install -e .
+  pip3 install -e .
   # add gunicorn and gevent for hosting
-  pip install gunicorn gevent
+  pip3 install gunicorn gevent
 
-config file for bepasty -- ``/home/bepasty/bepasty.conf``:
+Config file for bepasty -- ``/home/bepasty/bepasty.conf``:
 
 Copy ``src/bepasty/config.py`` to ``/home/bepasty/bepasty.conf`` first,
 remove the ``class Config`` and remove all indents in the file.
-The comments can be removed too, if you feel the need to.
-At last modify these two configs variables:
+The comments can be removed too, if you feel the need to.  If you wish 
+to use a script to do this in one go - please see the Tips & Tricks below.
+
+At least modify these four config variables.  
 
 ::
 
+  SITENAME = 'bepasty.example.org'
   STORAGE = 'filesystem'
   STORAGE_FILESYSTEM_DIRECTORY = '/home/bepasty/storage/'
+  SESSION_COOKIE_SECURE = False
 
-
-add this content to ``/home/bepasty/bin/gunicorn_bepasty``:
+Add this content to ``/home/bepasty/bin/gunicorn_bepasty``:
 
 ::
 
@@ -58,6 +76,7 @@ add this content to ``/home/bepasty/bin/gunicorn_bepasty``:
   SOCKFILE=$HOME/gunicorn.sock  # we will communicate using this unix socket
   PIDFILE=$HOME/gunicorn.pid
   NUM_WORKERS=3                 # how many worker processes should Gunicorn spawn
+  export PATH="/home/bepasty/.local/bin:$PATH"  # include gunicorn path for exec
   export BEPASTY_CONFIG=$HOME/bepasty.conf
 
   source $HOME/bin/activate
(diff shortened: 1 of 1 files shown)


docs/source/install-tutorial.rst
```diff
@@ -3,21 +3,36 @@
...
-  adduser bepasty
+  adduser bepasty --disabled-login
+  
+Commands to run as user bepasty:
+
+::
+
   # change to user bepasty
-  sudo su - bepasty
+  su - bepasty
```
ThomasWaldmann, line 31: your change here only works if the user bepasty has a password set. the sudo method should also work without that.

docs/source/install-tutorial.rst
```diff
@@ -3,21 +3,36 @@
...
+  # add path to bepasty/.bashrc
+  echo "export PATH=\"/home/bepasty/.local/bin:\$PATH\"" >> /home/bepasty/.bashrc
```
ThomasWaldmann, line 33: use ~/.bashrc here also, see below.

docs/source/install-tutorial.rst
```diff
@@ -30,24 +45,27 @@ commands to run
...
 ::
 
+  SITENAME = 'bepasty.example.org'
   STORAGE = 'filesystem'
   STORAGE_FILESYSTEM_DIRECTORY = '/home/bepasty/storage/'
+  SESSION_COOKIE_SECURE = False
 
-
-add this content to ``/home/bepasty/bin/gunicorn_bepasty``:
+Add this content to ``/home/bepasty/bin/gunicorn_bepasty``:
```
ThomasWaldmann, line 68: you added `.local/bin` to PATH above, but here you use `bin`.

docs/source/install-tutorial.rst
```diff
@@ -115,3 +138,43 @@ Supervisord config i.e. in ``/etc/supervisor/conf.d/bepasty.conf``:
...
+* When adding additional users in ``bepasty/config.py`` make sure you have
+  ``'user_secret': 'correct,list,of,permissions',``
+  any missing ``'' : '',`` will result in the bepasty server not starting.
+  
+Tips & Tricks:
+* To convert the ``src/bepasty/config.py`` to ``/home/bepasty/bepasty.conf`` in one easy go.
+  On the command line as the bepasty user:
+  # create trim alias
+  ``alias trim="awk '{\$1=\$1};1'"``
+  # remove the top 11 lines + trim spaces at beginning of lines
```
ThomasWaldmann, line 173: this looks like something that is going to break if we ever change the header of that file.

docs/source/install-tutorial.rst
```diff
@@ -115,3 +138,43 @@ Supervisord config i.e. in ``/etc/supervisor/conf.d/bepasty.conf``:
...
+  ``trim src/bepasty/config.py | tail -n +11 > /home/bepasty/bepasty.conf``
+* The user secret in ``bepasty/config.py`` needs to be secure - an easy way 
+  to create secure random character strings on the command line is:
+  ``tr -dc A-Za-z0-9 </dev/urandom | head -c 10 ; echo ''``
```
ThomasWaldmann, line 177: the user secret is something a user either interactively / manually enters at the web UI - in that case they will use something that they can remember more easily.

or, they use a password manager, in that case they won't need that script because they'll use the pwmgr to generated the pw.

docs/source/install-tutorial.rst
```diff
@@ -115,3 +138,43 @@ Supervisord config i.e. in ``/etc/supervisor/conf.d/bepasty.conf``:
...
+  this can also be used to create a long random secure SECRET_KEY value:
+  ``tr -dc 'A-Za-z0-9!"#$%&'\''()*+,-./:;<=>?@[\]^_`{|}~' </dev/urandom | head -c 255  ; echo ''``
+  The length of the random character string is set by the head -c $LENGTH.
```
ThomasWaldmann, line 180: yes, or you can just type a long enough more or less random string, which will be fine also.

you'll never need to enter it, so that's pretty easy without a complicated looking script piece like that.

so, can we just get rid of these pw generating tipps and tricks?

i'ld rather like to keep the docs shorter and simpler than make a shell scripting tips and tricks out of them.

docs/source/install-tutorial.rst
ThomasWaldmann: so, just remove that, please.

ThomasWaldmann:
@oijn did you see my feedback?

docs/source/install-tutorial.rst
oijn: sudo is part of the default install & config for Ubuntu but will need to be installed and configured for a Debian minimal server install.  I have no problems either way.

docs/source/install-tutorial.rst
oijn: Fair enough.  Do you have any suggestions on an alternative approach to automating this given the amount of manual effort needed to get the config file into a suitable format and the consequences and errors if not?

docs/source/install-tutorial.rst
oijn: The above was added to suppress errors from when installing flask &  pygmentize during the pip3 install.

docs/source/install-tutorial.rst
oijn: Don't quite follow?  
If you are referring to /home/bepasty/bin/gunicorn_bepasty - this came from the previous instructions. I didn't change anything here and gunicorn seems to expect the config there.

docs/source/install-tutorial.rst
oijn: I have no problems either way.

docs/source/install-tutorial.rst
oijn: I have no problems either way.

ThomasWaldmann: ok, so just remove these lines.

docs/source/install-tutorial.rst
ThomasWaldmann: same here.

docs/source/install-tutorial.rst
ThomasWaldmann: the install tutorial is for a manual installation, so it does not need to do everything automated.

docs/source/install-tutorial.rst
ThomasWaldmann: ehrm, ignore my comment. it applies for the case when one is **not** root. But at this point, the user **is** root (according to the updated instructions), so `sudo` is not needed.

ThomasWaldmann, line 31: strictly taken, line 26 needs to be moved to below current line 31.

docs/source/install-tutorial.rst
ThomasWaldmann: what I meant is that in line 35, you use `~/.bashrc` but in line 33 you use `>> /home/bepasty/.bashrc` to refer to the same file.

i suggest to use `>> ~/.bashrc` in line 33 also.

ThomasWaldmann: still wondering though: what exactly ends up being in `.local/bin` when following these instructions?

shouldn't all stuff end up in `./bin` of the virtual env we created and activated?

docs/source/install-tutorial.rst
ThomasWaldmann: ah, yeah, ignore my comment. this `bin` comes from the virtualenv.

ThomasWaldmann:
besides my feedback, one could also add that `pip install -e[pillow]` might be desired if one wants to show thumbnails.

see that PR: https://github.com/bepasty/bepasty-server/pull/271

PR #273, opened 2021-04-17
Document how to authenticate with the REST api

Duckle29:


docs/source/rest.rst
```diff
@@ -18,6 +18,9 @@ Currently the REST API provides the following API Endpoints::
     POST /apis/rest/items/<itemname>/unlock
 
 
+Authentication
+==============
+For endpoints that require non-default permissions authentication can be done through the http basic authentication, with the user ``user`` and the password for the intended permissions.
```
ThomasWaldmann, line 23: Thanks for adding this to the docs (even I forgot about how that would work). But it needs a little fix:

From the 0.3.0 changelog entry:

support http basic auth header (it just reads the password from there, **the user name is ignored**).

this is useful for scripting, e.g. you can do now: $ curl -F 'file=@somefile;type=text/plain' http://user:password@localhost:5000/+upload

Duckle29: Cool I made the change, do you want the example with curl to be added as well?

ThomasWaldmann: No, people can look that up in the docs of the tool they use.

ThomasWaldmann:
Thanks!

PR #274, opened 2021-06-25
Revert src layout

aug0stus:
Eliminating the src/ folder will allow direct import of the bepasty module from the project's root directory and permit the application to be served as so, `gunicorn bepasty.wsgi`.


aug0stus:
Whoops, this PR was meant to target my fork, hehe.

ThomasWaldmann:
if you need to do that, you're likely doing it wrong. :-)

ThomasWaldmann:
yeah, but you really want some sort of "pip install", i personally use a virtual env and `pip install -e .` .

ThomasWaldmann:
btw, the src/ layout is quite popular amongst quite some long-term python developers to avoid some weird issues which can bite you when not using it. so it is there for a reason.

aug0stus:
Yeah, I've seen the src/ layout in many places. It's even in the [setuptools](https://setuptools.readthedocs.io/en/latest/userguide/declarative_config.html#using-a-src-layout) documentation.

Yet it still seems that the user base is divided on it. I did search on a few popular python projects that I know of, and they followed the `<project name>/` layout. 

I've only managed to study two python-based projects so far, and they both have wildly different ways of implementing their containers. [Elasticsearch Curator](https://github.com/elastic/curator) is compiling the python code into an binary executable. 

I've been learning a lot the last few days.

PR #275, opened 2021-06-27
Use module instead of class for flask configuration

aug0stus:
Flask module supports module as objects for configuration abilities.

This will allow easier user overrides as they won't need to manipulate
the class into a data configuration file.


ThomasWaldmann (changes requested):
Thanks for the PR!

I agree that using a module simplifies setup and docs a lot.

But there are 2 issues with this PR:

- i don't really see the point of renaming all "config" to "settings" (including changing the module name of the builtin default config) - this makes your change unnecessarily big and also hard to review, so please remove that.
- if a working existing installation stops to work and requires manual changing, this is a breaking change and guess we would have to do a major release (which can be done, but the point in time has to be chosen wisely). In case this takes a while a changeset with widespread changes (like due to config -> settings change) is more likely to cause merge conflicts than a smaller change (only change from class to module).

src/bepasty/tests/test_rest_server.py
```diff
@@ -14,8 +14,8 @@
 
 import pytest
 
-from ..app import create_app
-from ..config import Config
+from bepasty.app import create_app
+from bepasty import config
```
ThomasWaldmann, line 18: please keep the relative imports.

ThomasWaldmann: the 2nd one would then be `from .. import config` I guess.

src/bepasty/config.py
```diff
@@ -1,136 +1,133 @@
...
+#:
+#: ::
+#:
+#:     PERMISSIONS = {
+#:         'myadminsecret_1.21d-3!wdar34': 'admin,list,create,modify,read,delete',
+#:         'uploadersecret_rtghtrbrrrfsd': 'create,read',
+#:         'joe_doe_89359299887711335537': 'create,read,delete',
+#:     }
+PERMISSIONS = {
+    # 'foo': 'admin,list,create,modify,read,delete'
```
ThomasWaldmann, line 128: please keep the line as it was, the comma at the end was intended. Just indent it correctly to match your other changes.

ThomasWaldmann (commented):
much better, thanks - just a few nitpicks to fix.

btw, appending ?w=1 to the url makes github PR diff view ignore whitespace, easier to review then.

src/bepasty/config.py
aug0stus: Done :). Neat trick with the ` ?w=1`. I couldn't figure it out at the beginning but I found a button in the GitHub UI page which hinted me to its functionality.

src/bepasty/tests/test_rest_server.py
aug0stus: I find the relative imports look a bit out of place, but I reverted anyways.

docs/source/quickstart.rst
```diff
@@ -27,18 +27,16 @@ especially when setting up the ``DEFAULT_PERMISSIONS`` (which apply to not-logge
 
 Here is the documentation straight from its config:
 
-.. autoclass:: bepasty.config.Config
+.. automodule:: bepasty.config
```
aug0stus, line 30: @ThomasWaldmann Could you have a look at the generated docs and share your opinion on the rendering from `.. automodule::` ? I find that it doesn't look too well integrated into the documentation. I liked the automatic `[source]` link and header stylistic provided by `.. autoclass::` but I don't think this is available for `.. automodule::`. 

To make it look a *little* bit better, there's a setting `add_module_names = True` in `conf.py` which removes the prefix from the attributes. It helps the aesthetics a little bit, and we could also move the disclaimer from the docstring to elsewhere.

```
IMPORTANT:

The config is only loaded at startup time of the app, so if you change it,
you need to restart the wsgi app process(es) to make it load the updated
config.
```

src/bepasty/app.py
```diff
@@ -69,7 +69,7 @@ def setup_secret_key(app):
 def create_app():
     app = Flask(__name__)
 
-    app.config.from_object('bepasty.config.Config')
+    app.config.from_object('bepasty.config')
     if os.environ.get('BEPASTY_CONFIG'):
         app.config.from_envvar('BEPASTY_CONFIG')
```
ThomasWaldmann, line 74: btw, did you practically try whether overriding a default config value (residing in `bepasty.config`) via a custom config module given by `BEPASTY_CONFIG` actually works correctly?

docs/source/quickstart.rst
ThomasWaldmann: True is the default, but guess it looks better with add_module_names = False - so please add that below the comment describing this in conf.py.

src/bepasty/app.py
aug0stus: Yep, just re-tested and I can see my changes on the site.

``` bash
cat bepasty.conf && BEPASTY_CONF=$(pwd)/bepasty.conf bepasty-server
# SITENAME = 'bepasty.local'
# DEFAULT_PERMISSIONS = 'admin,list,create,modify,read,delete'
#  * Starting bepasty server...
#  * Serving Flask app 'bepasty.app' (lazy loading)
#  * Environment: production
#    WARNING: This is a development server. Do not use it in a production deployment.
#   Use a production WSGI server instead.
#  * Debug mode: off
#  * Running on http://127.0.0.1:5000/ (Press CTRL+C to quit)
```

ThomasWaldmann: ok, great!

docs/source/quickstart.rst
aug0stus: OK finally, I don't think the attribute prefixes look so bad if the modules' own docstring is moved into the documentation.

I modified the documentation very slightly to accommodate a "key:value" type source code, and I think it looks pretty good.

What do you think of ffc0c205bcd588d2d4965f2b81140419c4459db4 ?

ThomasWaldmann: I commented there...

ThomasWaldmann:
"Here are the configuration defaults:" (note: singular configuration)

Also please edit conf.py and add add_module_names = False.

aug0stus:
> 
> 
> "Here are the configuration defaults:" (note: singular configuration)
> 
> Also please edit conf.py and add add_module_names = False.

My bad, I wasn't paying attention to the plural forms. I just made the requested modifications.

ThomasWaldmann (approved):

ThomasWaldmann:
OK, looks good now! Thanks.

ThomasWaldmann:
Assigned this to release 2.0.0 milestone because it is a breaking change and people need to edit their configs to keep it working.

So guess I'll make a 1.1.0 release first as soon as that milestone is completed, then merge this PR and then later release it as 2.0.0 (maybe together with more [breaking?] changes).

aug0stus:
> OK, looks good now! Thanks.

> Assigned this to release 2.0.0 milestone because it is a breaking change and people need to edit their configs to keep it working.
> 
> So guess I'll make a 1.1.0 release first as soon as that milestone is completed, then merge this PR and then later release it as 2.0.0 (maybe together with more [breaking?] changes).

:) thanks so much for your patience and opinions with the review process.

ThomasWaldmann:
@augustfengd Hmm, why did you delete this?

augustfengd:
Hey @ThomasWaldmann! I was cleaning up some inactive repositories that I owned. It must've closed this pull request.

ThomasWaldmann:
See #323.

PR #301, opened 2023-05-13
docs: REST POST response returns file URI in Content-Location

aither64:
While working on a [CLI in ruby](https://github.com/aither64/bepasty-client), I've discovered that the docs mention file URI in REST API being returned in `Content-Disposition`, while actually it is returned in `Content-Location`.

I'd also like to ask if you're ok with me naming the CLI bepasty-client? Would you merge a PR adding it to `docs/source/user-cli.rst`? I found pastebinit outdated and annoying to work with, so I've implemented something for my own purposes.

### DIFF
--- a/docs/source/rest.rst
+++ b/docs/source/rest.rst
@@ -142,7 +142,7 @@ POST Response by the server:
     *Transaction-ID*
         Transaction-ID provided for continued upload in a chunked upload
         process.
-    *Content-Disposition*
+    *Content-Location*
         The URI of the newly uploaded file on the server. Will only be
         provided when upload is finished and successful.
 


aither64:
Oh, of course I've overlooked https://github.com/bepasty/bepasty-client-cli. Well, since I've already wrote another... x)

ThomasWaldmann:
Yeah, that seems to be a documentation bug, thanks for the PR!

About your cli name: please name it so it won't be confused with the python client (package name as well as cli command name).

aither64:
Ok, I've renamed it to ruby-bepasty-client and the cli command is bepastyrb to keep it short. I hope that's all right with you, otherwise let me know :)

PR #303, opened 2023-05-19
utils/upload: guess file content type also using pygments

aither64:
It is possible for `mimetypes.guess_type()` to not recognize a file which is otherwise supported by pygments. It happens for example on Nix files (`*.nix`), which have `text/x-nix` mimetype. This patch makes bepasty to try to recognize a file using `pygments.lexers.get_lexer_for_filename()` if `mimetypes.guess_type()` does not know it.

### DIFF
--- a/src/bepasty/utils/upload.py
+++ b/src/bepasty/utils/upload.py
@@ -1,6 +1,8 @@
 import re
 import time
 import mimetypes
+from pygments.lexers import get_lexer_for_filename
+from pygments.util import ClassNotFound as NoPygmentsLexer
 from werkzeug.exceptions import BadRequest, RequestEntityTooLarge
 
 from flask import current_app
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Can you check if the failing test is related to your change or does it also happen without your change?
```
response = <WrapperTestResponse streamed [415 UNSUPPORTED MEDIA TYPE]>
code = 400, ftype = 'application/json', check_data = True
```

aither64:
I've tried to run it locally and I have the same failure on master on py310. It seems to be an error in the test. I'd say that the unsupported media type comes from here: https://github.com/pallets/werkzeug/blob/76b049dd45fd072fb62a54bccc0e8d513b03f4d8/src/werkzeug/wrappers/request.py#L654. See also the doc at https://github.com/pallets/werkzeug/blob/76b049dd45fd072fb62a54bccc0e8d513b03f4d8/src/werkzeug/wrappers/request.py#L556. It is called from https://github.com/bepasty/bepasty-server/blob/master/src/bepasty/apis/rest.py#L259. Apparently the 415 code is a new behaviour since v2.3... I've added a commit to fix it.

aither64:
Hm, the py37 environment has werkzeug v2.2, where it returns HTTP 400. So either the test would have to accept both, or I don't know how to go about it.

ThomasWaldmann:
see #304 and please rebase onto a current master after #304 is merged into master.

ThomasWaldmann:
Thanks for the PR!

PR #317, opened 2024-01-05
Fallback to Chrom(e|ium) if the Firefox driver for selenium isn't available

valholl:
Installing the selenium driver for Firefox under Debian is a bit of a pain, so when trying to run the tests locally I've found it easier to change the code to fallback to the Chromium driver if using the Firefox one fails.

As a Firefox user myself (for everything else) I would consider it extremely reasonable if this PR was refused, but I felt it was right to submit it in case it was useful for somebody else.

### DIFF
--- a/src/bepasty/tests/test_website.py
+++ b/src/bepasty/tests/test_website.py
@@ -1,6 +1,10 @@
 from selenium.webdriver import Firefox
+from selenium.webdriver.chrome.service import Service as ChromeService
+from selenium.webdriver import Chrome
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Thanks for the PR!

Good to have it if it makes life easier for some developers.

Chromium should be rather ok as it does not have the problematic components which Chrome has. But for automated testing use I guess they would not be that problematic anyway.

valholl:
Thanks!

PR #342, opened 2026-03-01
Move frontend from Bootstrap to Bulma and remove jQuery; dark theme support; some additional features

Merlin04:
Hi, and thanks for the wonderful project - I realize this is a pretty big PR to make without any prior discussion, so please don't feel any obligation to work on this or merge this if it wouldn't be helpful to bepasty! I made these changes for my own personal instance of bepasty (https://paste.tail917c8.ts.net) and I figured I might as well open a PR in case you all might want to merge them.

<details>
<summary>Screenshots</summary>
image
image
image
</details>

### Architectural changes
- Migrate frontend from Bootstrap to Bulma, removing jQuery along the way
  - This allows for dark theme, reduced network/JS payload, and removal of legacy dependencies
- Replace jQuery file uploader with [FilePond](https://pqina.nl/filepond/), an MIT-licensed async chunking file upload widget
- replace jQuery-based carousel with [fslightbox.js](https://fslightbox.com/)

### New features
- Dark theme! Uses the `prefers-color-scheme` CSS media query to detect user's preference. pygments uses the `monokai` theme when dark theme is active
- You can now click a button to instantly undo a binary file upload
screenshot of file upload widget showing a successfully uploaded file, with the text tap to undo

- New "Copy URLs" feature lets you copy a newline-separated list of urls to uploaded files (inline or preview). This is great for situations where the list feature is overkill, like if you're uploading a few media files to share with someone on a chat app
copy urls tab is open instead of create list tab. there's a selector for inline or preview, and below that a text field with the URLs that you can click to copy to clipboard

### Regressions/possible issues
- Some libraries are being loaded from a CDN because there isn't an xstatic package for them. I'm not super familiar with xstatic or the python packaging ecosystem in general, so maybe there's a better way to do this?
  - specifically, that's bulma.min.css, fslightbox.js, and filepond.js
- Carousel takes a slightly different format - it's now images on a page, and when you click one a full screen lightbox carousel view opens. I prefer this over the previous behavior, but I'm unsure if others would given how different it is from previous behavior
  - I've been thinking about possibly switching from fslightbox to lightgallery.js or photoswipe. Both of these libraries would provide a much nicer looking gallery view that puts thumbnails in a grid; however, they require knowing the resolution of the images, so I would have to add a function to the backend to calculate this (with similar logic to the thumbnail generator) 
- Use of some inline style tags in templates instead of a single CSS file when they only apply to UI components on a single page - I'm unsure if you have a preference for one or the other, but I didn't see many uses of inline style tags in the codebase so I figure I should bring this up
- I removed the documentation link from the navbar and adjusted the index page shown to logged-out users a bit to match my personal preference - I like it better this way but that's just personal preference

### WIP/TODOs
- Tests haven't been updated to match new element names (this is why the PR is a draft)
- `AUTHORS` and `CHANGES.rst` have not been updated

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,9 +36,7 @@ dependencies = [
     "xstatic < 2.0.0",
     "XStatic-asciinema-player <= 2.6.1.1",
     "xstatic-font-awesome >= 6.0.0, <= 6.2.1.1",
-    "xstatic-pygments <= 2.9.0.1",
-    "setuptools>=82.0.0",
-    "pillow>=12.1.1",
+    "xstatic-pygments <= 2.9.0.1"
 ]
 
 [project.scripts]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,13 +35,10 @@ dependencies = [
     "Pygments >= 2.12.0",
     "xstatic < 2.0.0",
     "XStatic-asciinema-player <= 2.6.1.1",
-    "xstatic-bootbox >= 5.4.0, <= 5.5.1.1",
-    "xstatic-bootstrap >=4.0.0.0, <= 4.5.3.1",
     "xstatic-font-awesome >= 6.0.0, <= 6.2.1.1",
-    "xstatic-jquery <= 3.5.1.1",
-    "xstatic-jquery-ui <= 1.13.0.1",
-    "xstatic-jquery-file-upload <= 10.31.0.1",
     "xstatic-pygments <= 2.9.0.1",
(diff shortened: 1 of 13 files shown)


Merlin04:
@ThomasWaldmann Hi, hope you're doing well! I have a bit more time now and might be able to work on this PR more, so I wanted to ask - are you interested in the possibility of merging this, or would you prefer it to remain a fork? If the former, I could try to finish up the remaining bits of the PR so it can exit draft stage.

